PR #4, opened 2024-01-11
Publish only when merging to master

kennsippell:


mrjones-plip (approved):
thanks kenn!  good call.

As well, we should see what how other repose that push to ECR protect against a build to a version that's already published (eg we have two or three versions of `1.0.0` on ECR now )

PR #13, opened 2024-01-19
move publish.js to script dir, delete dist dir

mrjones-plip:


package.json
```diff
@@ -43,9 +43,9 @@
     "cp-package-json": "cp package.json ./src",
     "test": "npx ts-mocha test/{,**}/*.spec.ts",
     "build": "npm run cp-package-json && npx tsc",
-    "start": "node dist/index.js",
+    "start": "node src/index.ts",
```
kennsippell, line 46: Do you really want to change this? I think index.js is still the build output from 'npm run build'

kennsippell: Confirmed that this should stay as-is
```suggestion
    "start": "node dist/index.ts",
```

kennsippell (approved):
Good to go when you revert that one liner

package.json
mrjones-plip: awesome - thanks for catching this!

mrjones-plip:
Thanks for the review Kenn, much obliged!

Issue #31, opened 2024-01-23
Support for DateTime (dob) property on contacts


kennsippell:
Uganda requires a DoB for contacts. Should have validation and datepicker

PR #41, opened 2024-01-26
Support for Date of Birth inputs

kennsippell:
#31

Watch base branch when merging

image

* Validation, Formatting, UI date picker using Bulma datepicker


kennsippell:
Feedback from @jonathanbataire 

> Dob should support just a number of years old and infer from today's date

freddieptf (approved):
LGTM. @jonathanbataire's feedback is valid and a simple number input reduces the clicks for the user but we sacrifice accuracy. I think the calendar widget should be fine and we can get more feedback from users after usage

kennsippell:
@jonathanbataire Thoughts?

jonathanbataire:
> @jonathanbataire Thoughts?

it should support both age as a number eg 37 and DoB in the `yyyy-mm-dd` format at least in the generated json
for age once someone submits a number (37) we can use today's date to calculate dob and save that dob in the json
support age because some people have no idea when their dob is but can tell their age

kennsippell:
Waiting for #40 before merging

PR #43, opened 2024-01-27
Fix env var table formatting in main readme.md

mrjones-plip:
The one with the 5 char change ([reference](https://en.wikipedia.org/wiki/List_of_Friends_episodes) (I'm totally not a Friends fan, but there you are))


kennsippell (approved):
cool

Issue #53, opened 2024-02-06
User blocked from downloading credentials due to cookie expiration


kennsippell:
1. User initiates a large upload
2. Waits a long time (hours) for upload to complete
3. Attempts to download credentials

Expected: Resulting credentials
Observed: Login page

This is because cookies expire after 1 day.

mrjones-plip:
Poking around a bit, I see we set the session cookie [only once at login](https://github.com/medic/cht-user-management/blob/79472e98cedef73ddc2839344ad6a3177a825654/src/routes/authentication.ts#L46). I think there's a holistic fix for the UX of "my session timed out and I didn't know it" which covers failed download of credentials in this ticket. Suggestion woudl be to renew the cookie to expire `LOGIN_EXPIRES_AFTER_MS` from time of current page load.  This way session expires `LOGIN_EXPIRES_AFTER_MS` after they last used the app instead of `LOGIN_EXPIRES_AFTER_MS` after they logged in.

Additionally, we could extend the app do some sort of `setInterval()` call where if the session cookie  is expired the user forcefully is redirected to a login page.  This covers the UX of being logged in, attempting to use the page rendered when logged in, but every AJAX request returns the [HTML of a login page](https://medic.slack.com/archives/C066J5C8Z88/p1707248364014699).

billwambua:
Also the suggestion to auto download credentials on upload completion by @mrjones-plip is also good.

Issue #18, opened 2024-01-20
Improve experience when working with large lists of users


kennsippell:
Currently things get a bit slow when working with a fully county's CHPs.

- [ ] #63
- [x] #64
- [ ] #61
- [ ] #62

kennsippell:
Closing in favour of more granular issues

PR #56, opened 2024-02-08
Add UI directive, progress bar, and status counts

kennsippell:
User is now greeted with a default "Getting Started" message
image

Once you have a user staged and ready for creation, the next step is clear
image

When user uploads, they see a progress bar
image

Once they upload, they are prompted to save the credentials
image

In a future change, the tags on the rigth side can be updated to be used as filters. https://github.com/medic/cht-user-management/issues/61

Progress bar in https://github.com/medic/cht-user-management/issues/18
In response to, but not a fix for https://github.com/medic/cht-user-management/issues/53
Closes https://github.com/medic/cht-user-management/issues/64


jonathanbataire:
good stuff
can we also add the upload all button to the main screen rather than having it under the drop down menu

jonathanbataire:
Demo'ed and tested these UI changes it looks great and work for me @freddieptf @kennsippell

src/public/place/directive.html
```diff
@@ -0,0 +1,29 @@
+<div class="columns" id="directive" hx-swap-oob="true">
+  <div class="column is-10">
+    <section class="hero box is-small is-light">
+      <div class="hero-body">
+        {% if progress.inProgressCount > 0 %}
+          {% include "place/directive_3_in_progress.html" %}
+        {% elsif progress.successCount > 0 %}
```
freddieptf, line 7: We should show the upload prompt anytime we have anything staged

src/public/place/directive.html
freddieptf, line 1: Can we have consistent header sizes and spacing with the views that already exist on the page? I also find the shadow/elevation on the boxes unnecessary

freddieptf (changes requested):

src/public/place/directive.html
kennsippell: > Can we have consistent header sizes and spacing with the views that already exist on the page? 

Sorry, I don't know what this means. Would you clarify?

kennsippell: > I also find the shadow/elevation on the boxes unnecessary

This is what the UI looks like without boxes. I don't particularly like this because nothing is aligned. I find there is no  line to the page. 
image

I'm currently using the default [box control](https://bulma.io/documentation/elements/box/) from Bulma. Can I ask what you would like to see? Are you thinking no boxes (above) or less elevated boxes? Happy to implement, but I'm really bad at design and need a little more input on a solution.

src/public/place/directive.html
kennsippell: In the nav bar or in this "directive" page?  If we are showing it all the time, I'm inclined to use the navbar.

src/public/place/directive.html
freddieptf: I meant the headers in the directives are super huge compared to the other headers on the page. I find the elevation weird because it's "higher" than the navbar and the two boxes don't share the same background color. It's not a blocker so we can keep it and get a designer to have a look

src/public/place/directive.html
freddieptf: In the directive, if we upload successfully we get the download prompt. If we stage new changes after the initial upload, I'd expect the download prompt to change to the upload prompt again

src/public/place/directive.html
kennsippell: Capturing in https://github.com/medic/cht-user-management/issues/67 to get a designer to look at this. For now I'll reduce header size, match background color, remove elevation, and send another screenshot.

kennsippell: image

kennsippell: image

src/public/place/directive.html
kennsippell: ok. so you don't want to show both at once.

what if you upload 300 and there are 3 failures? should they stilll see upload, download, or something else?

src/public/place/directive.html
jonathanbataire: it looks better the way it was before

src/public/place/directive.html
freddieptf: I think download with an option to retry failed works in that case

Issue #17, opened 2024-01-20
Errors during upload are not retried


kennsippell:
I've seen TIMEOUT errors and document update conflict errors so far - both of which would be nice to automatically retry

kennsippell:
also document update conflict

kennsippell:
503s

PR #58, opened 2024-02-11
Improvements to retry logic

kennsippell:
#17 

* Updates so all axios requests are retried under specific conditions (500s, timeouts, etc)
* Special considerations for retrying passwords (400) and update-conflicts (409s)
* Unit tests capturing some retry scenarios which were called out as important in #17 
* Unit tests confirm the harmonious retry behavior of upload-manager and axios-retry
* Simple test server to setup and try retry scenarios


src/lib/retry-logic.ts
```diff
@@ -0,0 +1,71 @@
...
+  retryCondition: (err: AxiosError) => {
+    const status = err.response?.status;
+    return (!status || status >= 500) && isRetryAllowed(err);
+  },
+  onRetry: (retryCount: number, error: AxiosError, requestConfig: AxiosRequestConfig) => {
+    console.log(`${requestConfig.url} failure. Retrying (${retryCount})`);
+  },
+};
+
+export async function retryOnUpdateConflict<T>(funcWithPut: () => Promise<T>): Promise<T> {
```
freddieptf, line 20: Is it possible to also handle this in the retry condition defined in the retry config?

src/lib/cht-api.ts
```diff
@@ -1,8 +1,12 @@
 import _ from 'lodash';
 import axios, { AxiosHeaders } from 'axios';
+import axiosRetry from 'axios-retry';
+import { axiosRetryConfig } from './retry-logic';
 import { UserPayload } from '../services/user-payload';
 import { AuthenticationInfo, Config, ContactType } from '../config';
 
+axiosRetry(axios, axiosRetryConfig);
```
freddieptf, line 8: Does this apply globally to the default axios instance? I think maybe we should now consider creating/configuring an axios instance for reuse within `ChtApi`

src/lib/retry-logic.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ChtApi } from './cht-api';
+
+const RETRY_COUNT = 4;
+
+export const axiosRetryConfig = {
+  retries: RETRY_COUNT,
+  retryDelay: () => 1000,
+  retryCondition: (err: AxiosError) => {
+    const status = err.response?.status;
+    return (!status || status >= 500) && isRetryAllowed(err);
```
freddieptf, line 13: Maybe we should directly target the status codes we know we can potentially recover from? Like the list in https://github.com/medic/cht-user-management/issues/17 503s, 409s

src/lib/retry-logic.ts
kennsippell: When an update conflict happens, the `rev` in the doc is outdated.  The correct way to resolve them is to GET, write, and PUT the object again.  If we just retry the PUT, we will either conflict again or overwrite somebody else's changes (in our case, sentinel).  This is why I handle separately.

src/lib/retry-logic.ts
kennsippell: Sure. 

`!status && isRetryAllowed(err)` I beleive is going to be mostly timeouts
`>= 500` can recover from 500, 502, 503, 504, 511.  As mentioned above, we can't handle 409s generally here.

src/lib/cht-api.ts
kennsippell: This adjusts the `axios` object within the `cht-api.ts` module. I don't think it is necessary to manage a singleton beyond what we are doing here. Is there a scenario you're concerned about?

kennsippell:
Thanks for the review @freddieptf.  Ready for 2nd round.

src/lib/cht-api.ts
freddieptf: We could tie it directly to the session in `ChtApi` setting the appropriate base url, request timeouts etc so it's cleaner to read/write. We can also eliminate much of the request config eg `this.authorizationOptions()` that we're currently doing

src/lib/retry-logic.ts
freddieptf: aah i see, we do need to do the GET first.

src/lib/cht-api.ts
kennsippell: Sorry, but I don't totally understand what you're imagining here. Can you clarify both the problem you're trying to solve and the approach you're envisioning?  Sounds like you are imagining a wrapper around axios?

freddieptf: Not a wrapper, just an instance of axios https://axios-http.com/docs/instance 
```
const instance = axios.create({
  baseURL: 'https://some-domain.com/api/',
  timeout: 1000,
  headers: {'X-Custom-Header': 'foobar'}
});
```

kennsippell: yes. i love this suggestion. can i do after https://github.com/medic/cht-user-management/pull/75 which has a refactor already to cht-api.ts ...  i think it'll just be an easier merge

jonathanbataire (approved):
Demo for UG

kennsippell:
@freddieptf Great feedback and I will address it. This is the base branch for #56. @jonathanbataire is demoing to Uganda and needs that today so I'm merging. This PR has been out for 2 weeks, so tons of time to give feedback. I will address the axiosInstance concept in https://github.com/medic/cht-user-management/pull/75.

PR #72, opened 2024-02-13
Deploy vLatest to production 

kennsippell:


freddieptf (approved):

package.json
```diff
@@ -54,7 +54,10 @@
     "lint": "npx eslint --color --cache .",
     "start": "node dist/index.js",
     "dev": "tsc-watch --onSuccess \"node dist/index.js\"",
-    "publish": "node scripts/publish.js"
+    "publish": "node scripts/publish.js",
+    "deploy-ke": "cd scripts/deploy && helm upgrade --kube-context arn:aws:eks:eu-west-2:720541322708:cluster/prod-cht-eks --namespace users-chis-prod --values values/users-chis-ke.yaml users-chis-ke medic/cht-user-management",
```
freddieptf, line 58: I'm not sure if this will work as expected, wouldn't this require having `medic/cht-user-management` in the deploy directory?

freddieptf: It might be worth replacing these vars with environment variables

kennsippell: I think when you run [these steps](https://github.com/medic/helm-charts/tree/main?tab=readme-ov-file#usage) then the `medic/cht-user-management` is an alias. I tested these commands and I don't have that repo in this folder. 

We can do environment variables, but I think just you and I are using this for the foreseeable future and then we are going to move to MoH self-host.  Did you try it? Can you?

kennsippell: Let's just leave it open. Can you try it next time you're doing a deployment? No rush on this one.

freddieptf:
I don't think we need this anymore. Closing this, updated deployment instructions in https://github.com/medic/cht-user-management/blob/main/scripts/deploy/medic-deploy.md

PR #73, opened 2024-02-14
Generate zip file with credential csvs

### LINKED ISSUE #21: Improvements to the export CSV with credentials
Currently there is no header on the exported credentials csv file which downloads. Would be nice to have a header

freddieptf:
Fixes #21
This PR changes the credentials file to a zip archive containing csv credential files for each level in the hierarchy that was uploaded

### DIFF
--- a/src/routes/files.ts
+++ b/src/routes/files.ts
@@ -29,8 +29,8 @@ export default async function files(fastify: FastifyInstance) {
       const record = [
         place.hierarchyProperties[parent.property_name],
         place.name,
-        place.contact.properties['name'],
-        place.contact.properties['phone'],
+        place.contact.properties.name,
+        place.contact.properties.phone,
         place.creationDetails.username,
         place.creationDetails.password,
       ];
--- a/src/routes/files.ts
+++ b/src/routes/files.ts
@@ -20,7 +20,7 @@ export default async function files(fastify: FastifyInstance) {
     return stringify([columns]);
   });
 
-  fastify.get('/files/credentials', async (req) => {
+  fastify.get('/files/credentials', async (req, reply) => {
     const sessionCache: SessionCache = req.sessionCache;
     const results = new Map<ContactType, String[][]>();
     const places = sessionCache.getPlaces();
@@ -37,7 +37,6 @@ export default async function files(fastify: FastifyInstance) {
       result.push(record);
       results.set(place.type, result);
     });
-
     const zip = new JSZip();
     results.forEach((places, contactType) => {
       const parent = Config.getParentProperty(contactType);
@@ -56,7 +55,7 @@ export default async function files(fastify: FastifyInstance) {
         })
       );
     });
-
+    reply.header('Content-Disposition', `attachment; filename="${Date.now()}_${req.chtSession.authInfo.friendly}_users.zip"`)
     return zip.generateNodeStream();
   });
 }
--- a/src/routes/files.ts
+++ b/src/routes/files.ts
@@ -1,9 +1,9 @@
 import _ from 'lodash';
 import { FastifyInstance } from 'fastify';
 import { transform, stringify } from 'csv/sync';
-import { Config } from '../config';
+import { Config, ContactType } from '../config';
 import SessionCache from '../services/session-cache';
-import Place from '../services/place';
+import JSZip from 'jszip';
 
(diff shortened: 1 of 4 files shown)


src/routes/files.ts
```diff
@@ -22,17 +22,41 @@ export default async function files(fastify: FastifyInstance) {
...
     const places = sessionCache.getPlaces();
-    const refinedRecords = transform(places, (place: Place) => {
-      return [
-        place.type.friendly,
+    places.forEach(place => {
+      const parent = Config.getParentProperty(place.type);
+      const record = [
+        place.hierarchyProperties[parent.property_name],
         place.name,
         place.creationDetails.username,
```
freddieptf, line 32: @kennsippell what do you think about adding the contact's name and phone number to the list? Would that be enough for cases where places share names?

kennsippell: I'd probably prioritize having the parent's name (eg CHU or sub-county). That seems super useful for sorting, distribution, and other activities.

I dunno if the contact's name and phone number seems valuable. I have no objections to adding it.

src/routes/files.ts
```diff
@@ -22,17 +22,41 @@ export default async function files(fastify: FastifyInstance) {
...
         place.creationDetails.password,
         place.creationDetails.disabledUsers,
       ];
+      const result = results.get(place.type) || [];
```
kennsippell, line 36: can't we move this to line 28? do it once outside the loop

src/routes/files.ts
```diff
@@ -22,17 +22,41 @@ export default async function files(fastify: FastifyInstance) {
...
+      const parent = Config.getParentProperty(contactType);
+      const columns = [
+        parent.friendly_name,
+        contactType.friendly,
+        'username',
+        'password',
+        'disabledUsers',
+      ];
+      zip.file(
+        `${contactType.name}.csv`,
```
kennsippell, line 52: i would love to see current timestamp in the filename to help disambiguate them

src/routes/files.ts
freddieptf: It has to be inside the loop, we're building the rows for each place's type

src/routes/files.ts
freddieptf: Yes i include the place's parent as the first column in the credential list. The problem is for the CHP lists, the users expect both CHU and Subcounty in the credential list and i was trying to give them more ways to identify the users

src/routes/files.ts
kennsippell: my mistake. it is the repetitive call to `set` that isn't needed. 
is the Map required? did you consider just using an object? believe you can do it in one line

something like: 
```
const results: { [key: string]: Array<string> } = {};
results[place.type] = [...results[place.type], record];
```

kennsippell (approved):
Nice work. A bit of polish and you can commit. Remember to `npm version patch` before commit.

src/routes/files.ts
freddieptf: I'd be using the object as a Map so i just used Map.

`results[place.type] = [...results[place.type], record];`

I think we'd be copying the array each time so while short, it's likely it's less efficient

src/routes/files.ts
freddieptf: > I don't think it is always on the contact.

I thought `phone` was only available on the contact model? Do you know when this isn't the case?

src/routes/files.ts
kennsippell: i believe spread operator performs faster in benchmarks ([example](https://stackoverflow.com/questions/39716216/difference-between-using-a-spread-syntax-and-push-apply-when-dealing-with))

the spread operator isn't a copy... and most of this is triviall optimized by a compiler 
overall, you should optimize for human brain space. that means prioritising readable code as the top priority rather than over optimizing small things

src/routes/files.ts
kennsippell: For Kenya phone is only on the contact, ya.  But it isn't required, and you can put the phone attribute on the place just as easily. 

It is just like any other attribute, so why assume it is only on the contact?

PhilipNgari:
@kennsippell  has this been merged to prod?

kennsippell:
@PhilipNgari Yep. In production.

Issue #11, opened 2024-01-19
Uganda - Restrict user creation to levels in the hierarchy (eg. at facility, district, country)


kennsippell:
>> It should support operations at different levels in the country ie at facility, district and country levels. This also means that the system restricts users by level of operation. Eg a user at facility should only be able to create only users for that facility

> Since the CHT doesn't support this this user-management-tool currently handles all authentication/authorization via CHT.  Are you okay if we rely on UI sugar to restrict how/where users can be created?
> Basically - a user could login to the CHT and create a user in a level outside their level. But this tool would only be able to create it a user within their level.

this should be okay

PR #75, opened 2024-02-15
Non-admin users can only create users under their user's place

kennsippell:
#11 

Non-admin users in the CHT are linked to a place and a contact ([see facility_id in schema](https://docs.communityhealthtoolkit.org/core/overview/db-schema/#users)). With the addition of custom [user management roles](https://github.com/moh-kenya/config-echis-2.0/pull/1808), we can now make many user-manager accounts with each one linked to a single place and contact.

After this change, these non-admin users:

1. Can only create users at or below their place in the CHT hierarchy
2. Can only see search results which are under their place in the CHT hierarchy
3. Must be linked to a place or they cannot login to CHT user management tool

Note that this is UI sugar and does not actually limit the permissions of these CHT user accounts. Users witjh these roles are able to create users anywhere on the instance (including other _user manager user_) via the CHT API. This is security through obfuscation.  Core feature request https://github.com/medic/cht-core/issues/8876

fyi @jonathanbataire @freddieptf


kennsippell:
Ping @inromualdo now that you're back from meetup

src/lib/cht-session.ts
```diff
@@ -0,0 +1,102 @@
...
+    return !!this.facilityId &&
+      (
+        this.facilityId === ADMIN_FACILITY_ID 
+        || remotePlace?.lineage?.includes(this.facilityId)
+        || remotePlace?.id === this.facilityId
+      );
+  }
+}
+
+async function createSessionToken(authInfo: AuthenticationInfo, username: string, password: string): Promise<string> {
```
inromualdo, line 62: Would it be better to refactor this function,  `fetchUserDetails` and `createUrl` as a private static method within `ChtSession` as there are used only in that function?

src/routes/authentication.ts
```diff
@@ -31,17 +31,17 @@ export default async function authentication(fastify: FastifyInstance) {
...
 
     const authInfo = Config.getAuthenticationInfo(domain);
-    let session;
+    let chtSession;
     try {
-      session = await ChtApi.createSession(authInfo, username, password);
+      chtSession = await ChtSession.create(authInfo, username, password);
     } catch (e: any) {
       return resp.view('src/liquid/auth/authentication_form.html', {
         domains: Config.getDomains,
```
inromualdo, line 39:
```suggestion
        domains: Config.getDomains(),
``` ?

File: src/types/fastify/index.d.ts [hunk 3]
```diff
@@ -1,4 +1,4 @@
-import { ChtSession } from '../../lib/cht-api';
+import ChtSession from '../../lib/cht-api';
```
inromualdo, line 1:
```suggestion
import ChtSession from '../../lib/cht-session';
```

inromualdo (changes requested):
Nice work @kennsippell . I added some comments/questions.

I would also recommend installing `npm install --save-dev @types/chai-as-promised` useful for ts to provide type checking

src/routes/authentication.ts
kennsippell: ... :P

src/lib/cht-session.ts
kennsippell: I'm not sure if I prefer having the function private because it is in the module scope or because it is explicitly private in the class scope... but I'm happy to update as you have suggested.

kennsippell:
Thanks for the great review @inromualdo

inromualdo (approved):

Issue #76, opened 2024-02-16
Non-admin users cannot safely find users by facility_id on > cht-core v4.4


kennsippell:
This tool needs to find a list of users by their `facility_id`.  To "replace" an existing user, we need to find the users which are linked to a place. Currently, we do this with [disableUsersByPlace](https://github.com/medic/cht-user-management/blob/main/src/lib/cht-api.ts#L141) but this isn't working for non-admin users.

Three issues combine to make this issue quick sticky:
* Non-admin users (users with `mm-online` role) get a 403 unauthorized when running the `_users/_find` query currently in use.  Access to the `_users` database is for admins only after CHT 4.4's upgrade of CouchDB.
* The `GET /api/v2/users` API endpoint doesn't support looking up users by facility_id https://github.com/medic/cht-core/issues/8877
* It isn't safe to run `/api/v2/users` to fetch all users and filter it in memory by facility_id https://github.com/medic/cht-core/issues/8689

This blocks "replace user" scenarios for devolved user management. cc @freddieptf

kennsippell:
Options?

1. Upgrade to 4.6
2. Backport fix for https://github.com/medic/cht-core/issues/8689 back to 4.2.x and upgrade
3. Modify `_users/_security` doc to permit access to the `user_manager` role

This only affects CHT-Core 4.4 or later. So we have a bit of time for this since eCHIS is on 4.2 and 4.3.

mrjones-plip:
> This could potentially block CHIS upgrades

thanks for the heads up!  Let's see if the [work around proposed](https://github.com/medic/cht-core/issues/8877#issuecomment-1948531882) in the ticket is viable and go from there. 

I'll be keeping an eye out for updates!

mrjones-plip:
Great to see the ddoc work around is deployed!

Note that  https://github.com/medic/cht-core/issues/8877 is under way and will be released in CHT `4.7.0`

kennsippell:
Reactivating due to https://github.com/medic/cht-user-management/issues/115

mrjones-plip:
@kennsippell - are you able to reproduce this in a dev environment?  I wonder if using the [new users API](https://github.com/medic/cht-core/issues/8877) going to be released in 4.7 would be a solution?  If you can reproduce it in a dev environment, then we could try using the [branch images](https://github.com/medic/cht-core/pull/8928) on `8877-lookup-single-user` to see if the new API fixes the timeout.  

[...]

PR #77, opened 2024-02-19
Use user-settings docs for disableUsersWithPlace

kennsippell:
To test this, replace a user on eCHIS running CHT-core >4.4 logged in via [this user-manager role](https://github.com/moh-kenya/config-echis-2.0/pull/1808).

#76


kennsippell:
Ping @alexosugo now that you're back from meetup

alexosugo:
@kennsippell I've experienced some challenges testing this with the training instance due to multiple duplicate CHUs. I'm unable to proceed beyond filling the replacement form.

kennsippell:
@alexosugo Training instances are running `4.2.2`, I believe. This will require `>4.4` to test, so training might not be the best place to test it. 

Also, I know you were having problems with login. This may have been related to https://github.com/medic/cht-user-management/issues/97 which is now fixed. I encourage you to test again and I would be happy to do a pair session with you if you get stuck.

alexosugo (approved):
Sorry it took a while. Tested and works great!

Issue #82, opened 2024-02-26
 Regex Validator Incorrectly Alters Email Addresses


paulpascal:
## Description

The regex validator in `validator-regex.ts` is currently altering email addresses in a way that makes them invalid. 
This is due to the format on the input (`validator-string.ts` format), which replaces certain characters (including `.` and `@`) with spaces or removes them entirely.

## Steps to Reproduce

1. Add a contact property of type `regex` for email attribute:
```
"contact_properties": [
        {
          "friendly_name": "Email address",
          "property_name": "email",
          "type": "regex",
          "parameter": "^[a-zA-Z0-9._+-]+@[a-zA-Z0-9.-]+\\.[a-zA-Z]{2,}$",
          "required": true
        },
],
```
  
2. Upload or add new item and define email attribute. Eg: `johndoe@gmail.com`

3. The validator alters the email address to `johndoegmail com`, which is not a valid email address.

## Expected Behavior

The regex validator should validate the email address, trim space around and retain all original characters, including `.` and `@`.

## Actual Behavior

The regex validator alters the email address, replacing `.` with a space and removing `@`.

<img width="1048" alt="Screenshot 2024-02-27 at 03 37 53" src="https://github.com/medic/cht-user-management/assets/16521690/8e0fa82f-0083-46c5-8e10-4e427d582eea">
<img width="1048" alt="Screenshot 2024-02-27 at 03 34 21" src="https://github.com/medic/cht-user-management/assets/16521690/bd043690-9930-47c7-b3bd-e7b7bef796c4">

## Suggested Fix

- Have a dedicated email validator
- Fix the regex validator

PR #85, opened 2024-03-08
Support for email inputs

kennsippell:
#82 

The ValidatorRegex uses ValidatorString. ValidatorString should control restrictions which we want to apply to all inputs. I'm relaxing that logic to allow email characters + moving some reformatting into ValidatorName where it is more appropriate.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cht-user-management",
-  "version": "1.0.15",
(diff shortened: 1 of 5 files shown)


paulpascal (approved):
Awesome @kennsippell 
LGTM✨

kennsippell:
Thanks for the quick review @paulpascal

Issue #55, opened 2024-02-08
Gender form input should be a dropdown instead of freetext field 


kennsippell:
> Sex should be a drop down with just male and female

https://github.com/medic/cht-user-management/pull/40#issuecomment-1931829645

inromualdo:
I propose removing this type and replacing it with a select_one type. This not only addresses the gender issue but also enables the selected items to be transformed into a dropdown. The parameter will be an object with key = select-option value and value = label.

@kennsippell

kennsippell:
Sounds good @inromualdo!  Interested to hear which of [these tests](https://github.com/medic/cht-user-management/blob/main/test/lib/validation.spec.ts#L55-L60) will work after the change. If we break a few, I think that'd be okay. 

This may also be useful for @paulpascal to address [this feedback](https://github.com/medic/cht-user-management/pull/86#discussion_r1519206120).

inromualdo:
It could just be `select`. I initially used `select_one`, thinking we might add `select_multiple` later. However, in the contact creation forms across several projects, I haven't seen any multiple-choice elements. So, we can simply stick with `select`.

It would look like this:
<img width="588" alt="Screenshot 2024-03-12 at 2 19 22 PM" src="https://github.com/medic/cht-user-management/assets/12269482/ab51b0ba-432e-4dbe-929a-d804c416244e">

Either options are added directly in the configs (Gender) or in a separate CSV file (Level Of Study).

<img width="376" alt="Screenshot 2024-03-12 at 2 27 54 PM" src="https://github.com/medic/cht-user-management/assets/12269482/9c02f219-3b31-4f59-9895-a331fe9a3a9f">

After the change, all gender tests will fail unless we leave the type `gender` and add the `select` type alongside it.

The genarated html will be:

<img width="440" alt="Screenshot 2024-03-12 at 2 36 16 PM" src="https://github.com/medic/cht-user-management/assets/12269482/38e31a57-3f50-4770-9b4d-54e604811d5c">

paulpascal:
@inromualdo , we will have `select multiple` with `roles` (related #86 )

inromualdo:
> How many levels of study are there?

This is just an example. Here, there is 5. But in the CIV project we have a select_one with more than 200 items.

> parameter already supports `array[]`

In this example, I use object, so that I can display a friendly label

> I doubt those would be easily handled by this `select_one` type

Those would be handled.

PR #86, opened 2024-03-11
Support multiple roles on a contact type

### LINKED ISSUE #69: Support for two user_roles (togo chw and rcom) at village level

paulpascal:
Fix #69 

- change `user_role` to `string[]`
- have a config function to generate `userRoleProperty` anytime we may need it (default property_name, type,... and only passing the `user_role` as `parameter`)
- implement changes in `Place` and `PlaceFactory` to process the new property
- add `ValidatorRole` to handle check `selected roles` against `user_role` (behaving as `allowed roles` when having more than one item)
- update in `UserPayload` (to only use `role` and no more `type`)
- update concerned templates 
- update doc


src/public/place/create_form.html
kennsippell, line 1: sorry - but i just merged a change that caused some conflicts here

test/services/upload-manager.spec.ts
```diff
@@ -245,6 +245,38 @@ describe('upload-manager.ts', () => {
     expect(chtApi.createUser.callCount).to.be.gt(2); // retried
     expect(chu.uploadError).to.include('could not create user');  
   });
+
+  it('mock data is properly sent to chtApi (multiple roles)', async () => {
```
kennsippell, line 249: would love to see a targeted test in `place.spec.ts` also targeting the serialization of form data into the place object

README.md
```diff
@@ -44,6 +44,7 @@ Property | Type | Description
 `contact_types.replacement_property` | Property | Defines how this `contact_type` is described when being replaced. The `property_name` is always `replacement`. See [ConfigProperty](#ConfigProperty).
 `contact_types.place_properties` | Array<ConfigProperty> | Defines the attributes which are collected and set on the user's created place. See [ConfigProperty](#ConfigProperty).
 `contact_types.contact_properties` | Array<ConfigProperty> | Defines the attributes which are collected and set on the user's primary contact doc. See [ConfigProperty](#ConfigProperty).
+`contact_types.user_roles_property` | ConfigProperty | Defines the property used for specifying user role(s) when creating users of this contact type. Add flexibility to choose role(s) when there are multiple roles available at this level of your hierarchy. See [ConfigProperty](#ConfigProperty).
```
kennsippell, line 47: I think it is important that we somehow capture the "allowed roles" at each level. Currently, this feature supports free-text input but what if somebody uses the wrong case? What if they use a _CHW role_ for a _supervisor user_ (this can actually result in quite serious bugs)? What if there are trailing spaces on the role, which is super common when working with CSVs?

Ideally, I think the best user experience for this feature would be a set of checkboxes like in the CHT today (see screenshot).  You don't need to implement that UI here in this PR, we can start with free-text; but if we capture the "allowed roles" and validate them, I think that's a good start.

image

There are probably a few ways to approach this.  But you'll want to validate this propery like the other property types.  You'll probably end up with something like `argument: ["role1", "role2"]`.

src/config/index.ts
```diff
@@ -23,6 +23,7 @@ export type ContactType = {
   replacement_property: ContactProperty;
   place_properties: ContactProperty[];
   contact_properties: ContactProperty[];
+  user_roles_property?: ContactProperty;
```
kennsippell, line 26: No objection to using the `ContactProperty` for this; but it also doesn't seem like the _best_ fit. Here are a few things I'm finding awkward with the interface.

1. `required` attribute is redundant because we have `user_roles_property` can be undefined
2. `property_name` is redundant because this can only control the user's role
3. `type` is not needed because this isn't validated
4. It is a bit confusing having both `user_role` and `user_roles_property`.  Could we somehow get to one?

This sort of awkwardness exists elsewhere in the config like `replacement_property`. Not sure how much work it would be to make a custom type for this; but just mentioning it.

src/services/user-payload.ts
```diff
@@ -29,6 +31,21 @@ export class UserPayload {
...
+   * If roles are present, this method returns an empty string.
+   * If roles are not present, it returns the user type defined.
+   * 
+   * When the type is sent alongside roles to CHT, roles are ignored.
+   * In such cases, this method returns an empty string to prioritize roles.
+   *
+   * @param place - The place object containing creation details and type information.
+   * @returns A string representing the user type.
+   */
+  public getUserType(place: Place): string {
```
kennsippell, line 45: i find this flow a bit confusing... it's kinda awkward that roles is sometimes in control and sometimes it is type. 

could we just always set `roles` and never set `type`?

kennsippell (changes requested):
Nice work, buddy! Really really nice start on this work.

src/public/place/create_form.html
```diff
@@ -29,6 +29,13 @@
       %}
     {% endfor %}
   </section>
+  <section class="section is-small">
+    {% if contactType.user_roles_property != null %}
```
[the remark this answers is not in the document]
paulpascal: Sure!

README.md
kennsippell: [This conversation may be useful](https://github.com/medic/cht-user-management/issues/55#issuecomment-1991383283)

src/config/index.ts
paulpascal: Thanks!

paulpascal:
Hi @kennsippell, thank you for your thorough review - I have worked on them. 
Whenever you have the time, could you please take another look? I appreciate your feedback. 
Thanks!

test/mocks.ts
```diff
@@ -127,3 +127,49 @@ export function expectInvalidProperties(
...
+  };
+};
+
+export const mockSimpleMultipleRolesContactType = (
+  propertyType,
+  propertyValidator: string | string[] | undefined,
+  errorDescription?: string,
+  allowedRoles?: string[],
+) : ContactType => {
+  const mockedProperty = mockProperty(propertyType, propertyValidator);
```
kennsippell, line 153: could this call `mockSimpleContactType` and just set `.user_role` after?

test/services/place.spec.ts
```diff
@@ -135,4 +135,34 @@ describe('service place.ts', () => {
     expect(actual.contact.type).to.eq(contactType.contact_type);
     expect(actual.contact.contact_type).to.be.undefined;
   });
+
+  it('setPropertiesFromFormData (support multiple roles)', () => {
```
kennsippell, line 139:
```suggestion
  it('setPropertiesFromFormData supports multiple roles', () => {
```

test/mocks.ts
```diff
@@ -127,3 +127,49 @@ export function expectInvalidProperties(
...
+  mockedProperty.errorDescription = errorDescription;
+  return {
+    name: 'contacttype-name',
+    friendly: 'friendly',
+    contact_type: 'contact-type',
+    user_role: mockUserRole(allowedRoles ?? ['role1', 'role2']),
```
kennsippell, line 159: better to use default parameters on line 51 instead of providing a default here?  `allowedRoles: string[] = ['role1', 'role2']` ?

src/liquid/place/bulk_create_form.html
```diff
@@ -79,6 +79,15 @@
               <td>Property on primary contact</td>
             </tr>
             {% endfor %}
+
+            {% if contactType.user_role.parameter != null %}
+            <tr>
+              <td>{{ contactType.user_role.friendly_name }}</td>
+              <td>{{ contactType.user_role.required }}</td>
+              <td>Role(s) of the user, use when there are multiple roles available at this level of your hierarchy. <br> <i>eg: rcom or supervisor+stock_manager </i> <br> (separate multiple roles with <b>+</b> or leave blank to use the default role of this level.)
```
kennsippell, line 87:
```suggestion
              <td>Role(s) of the user. Separate multiple roles with <b>+</b> or leave blank to use the default role <i>eg: rcom or supervisor+stock_manager </i>
```

I don't think we should say "when there are multiple roles available" because this only shows when there are multiple roles available?

test/lib/validation.spec.ts
```diff
@@ -133,5 +133,23 @@ describe('lib/validation', () => {
       description: `Cannot find 'contacttype-name' matching 'Sin Bad' under 'Parent'`,
     }]);
   });
+
+  it('user role is invalid when not allowed', () => {
```
kennsippell, line 137: i am currently able to create users with no role. can that be disallowed?

src/liquid/place/list.html
```diff
@@ -18,6 +18,9 @@ <h2 class="title is-4">{{contactType.friendly}}</h2>
             {% for contact_property in contactType.contact_properties %}
               <th id="contact_{{contact_property.property_name}}">{{contact_property.friendly_name}}</th>
             {% endfor %}
+            {% if contactType.user_role.allowed_roles != null %}
```
kennsippell, line 21: what is `user_role.allowed_roles`? I don't see the header working as expected

kennsippell (changes requested):
Great progress @paulpascal !

README.md
paulpascal: Tried to use it, but since `select_multiple` only accept `parameter` to be object for the moment (not simple array).

Got a simple `role validator` that can do the job, maybe the time to complete `select_multiple` feature.

paulpascal:
Hello @kennsippell, had some work after our discussion, and made changes accordingly.
Would you please review it back ?

Many thanks

README.md
```diff
@@ -37,7 +37,7 @@ Property | Type | Description
 `contact_types.name` | string | The name of the contact_type as it [appears in the app's base_settings.json](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/hierarchy/)
 `contact_types.friendly` | string | Friendly name of the contact type
 `contact_types.contact_type` | string | The contact_type of the primary contact. [As defined in base_settings.json](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/hierarchy/)
-`contact_types.user_role` | string | The [role](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/user-roles/) of the user which is created
+`contact_types.user_role` | string[] | The [role](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/user-roles/) to assign to the created user. When multiple roles are available, this property acts as a list of allowed roles against which the specified role(s) should be checked.
```
kennsippell, line 40:
```suggestion
`contact_types.user_role` | string[] | A list of allowed [user roles](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/user-roles/). If only one is provided, it will be assigned automatically.
```

src/lib/validator-role.ts
```diff
@@ -0,0 +1,33 @@
+import _ from 'lodash';
+import { ContactProperty } from '../config';
+import { IValidator } from './validation';
+
+export default class ValidatorRole implements IValidator {
```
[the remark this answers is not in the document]
paulpascal: Yes @kennsippell , of course. Was waiting for his work to be validated to rebase. 
But I can rebase right now it is okay with you.

paulpascal:
> Nice. Feels very close!

Thanks @kennsippell , feedback addressed

src/config/index.ts
```diff
@@ -102,6 +102,23 @@ export class Config {
...
+    return {
+      friendly_name: 'Role(s)',
+      property_name: 'role',
+      type: 'select_role',
+      required: true,
+      parameter: contactType.user_role,
+    };
+  }
+
+  public static supportsMultipleRoles(contactType: ContactType): boolean {
```
kennsippell, line 115:
```suggestion
  public static hasMultipleRoles(contactType: ContactType): boolean {
```

src/lib/validator-role.ts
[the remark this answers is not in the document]
kennsippell: ideally i think we would set Ro's work to be the base branch for this work. then we would merge Ro's PR first.

given the circumstances, i'm fine with pushing this change forward as-is and we can ensure we delete it in Ro's PR

src/liquid/place/bulk_create_form.html
```diff
@@ -79,6 +79,15 @@
               <td>Property on primary contact</td>
             </tr>
             {% endfor %}
+
+            {% if contactType.user_role.size > 1 %}
+            <tr>
+              <td>{{ userRoleProperty.friendly_name }}</td>
+              <td>{{ userRoleProperty.required }}</td>
+              <td>Role of the user. Specify multiple roles with <i>space</i>. Allowed roles are <i>{{ contactType.user_role | join: ', ' }}
```
kennsippell, line 87:
```suggestion
              <td>Role of the user. Separate multiple roles with a <i>space</i>. Allowed roles are <i>{{ contactType.user_role | join: ', ' }}
```

Can you have a `space` in a valid CouchDB role?  If so, we should change to `,`.

src/services/place.ts
```diff
@@ -216,6 +237,16 @@ export default class Place {
     return username;
   }
 
+  public extractUserRoles(): string[] {
```
kennsippell, line 240: should this just be a getter? `place.userRoles`?

test/services/place.spec.ts
```diff
@@ -135,4 +135,35 @@ describe('services/place.ts', () => {
     expect(actual.contact.type).to.eq(contactType.contact_type);
     expect(actual.contact.contact_type).to.be.undefined;
   });
+
+  it('setPropertiesFromFormData (supports multiple roles)', () => {
```
kennsippell, line 139:
```suggestion
  it('setPropertiesFromFormData supports multiple roles', () => {
```

test/services/place.spec.ts
```diff
@@ -135,4 +135,35 @@ describe('services/place.ts', () => {
...
+
+    const formData = {
+      place_prop: 'abc',
+      contact_prop: 'efg',
+      garbage: 'ghj',
+      user_role: 'role1 role2',
+    };
+    place.setPropertiesFromFormData(formData);
+
+    expect(place.properties).to.deep.eq({
```
kennsippell, line 154: nit: dunno if these assertions on place.properties are adding value since they are asserted elsewhere .up to you

kennsippell (approved):
Very well done, sir

Issue #48, opened 2024-01-30
CHP replacement - Hide/delete all information for outgoing CHPs instead of keeping the contact


kennsippell:
> as the new CHP, this is very confusing. Is it possible to hide the 'name of the old CHP' from the UI using a similar approach to the way we hide names of the deceased HHs members under the Deceased Card?

> I agree with 

> 1 on @katanu response. It is better to hide the name of the old CHW.

https://github.com/moh-kenya/config-echis-2.0/issues/1858#issuecomment-1916115050

jonathanbataire:
@kennsippell 
How are user credentials handled when replacing
for UG deployement MoH requires that no user credentials should be deleted 
so  we have a user role - "deactivated" that we assign to users we want to remove from the system the role has no permissions attached to it

kennsippell:
@jonathanbataire Super interesting. Thanks for sharing. 
Can I ask the goal of this?

Currently, the tool does the same thing that happens when you click "delete" (this red button in CHT). 
![image](https://github.com/medic/cht-user-management/assets/9014751/8525c782-1163-434c-954f-1269039e19d1)

This does the following:
1. Deletes the `user` doc in `_users` database
2. Does not delete the `user-settings` doc in `medic` database
3. Adds `inactive: true` flag onto `user-settings` doc 

So the user isn't currently "deleted" per se.  Would this work for Uganda users?

jonathanbataire:
> @jonathanbataire Super interesting. Thanks for sharing.
> 
[...]

PR #90, opened 2024-03-11
Replaced users in Uganda should be deactivated and not deleted

kennsippell:
#48

Replaced users in Uganda should be "deactivated" instead of deleted. This adds a new boolean flag `deactivate_users_on_replace` which toggles this behavior when set to true.

Note base branch here is from #89


paulpascal (approved):
Nice! @kennsippell 
Its all good to me

Issue #47, opened 2024-01-30
Autogenerate the VHT Area Name from VHT Name. Same for CHP Area Name from CHP Name


kennsippell:
> yes area name should be this Jonathan Bataire Area (village) and should be auto generated from name and village
> since area name is generated from vht name it should come after vht name.
https://github.com/medic/cht-user-management/issues/39#issuecomment-1914065651

kennsippell:
Togo also wants an autogenerated `name` and to provide `first_name` `last_name`

kennsippell:
Requirement for Togo: 

> 2 or more villages with the same name. Can we associate the name of the village with the name of its parent to facilitate the search

kennsippell:
Hey @inromualdo, since you offered to take this on - I thought I'd share some quick thoughts.

I'm imagining config.json like this, where `parameter` is a `liquidjs` template hydrated with some data like `place, contact, parent`.

Kenya
```
        {
          "friendly_name": "CHP Area Name",
          "property_name": "name",
          "type": "generated",
          "parameter": ["{{ place.name }} Area"],
          "required": true
        }
```

Togo
```
        {
          "friendly_name": "CHW Name",
          "property_name": "name",
          "type": "generated",
          "parameter": ["{{ place.first_name }} {{ place.last_name }}"],
          "required": true
        }
```

Uganda
```
        {
          "friendly_name": "VHT Area Name",
          "property_name": "name",
          "type": "generated",
          "parameter": ["{{ contact.name }} Area ({{ place.village }})"],
          "required": true
        }
```

Togo
```
        {
          "friendly_name": "Village Name",
          "property_name": "name",
          "type": "generated",
          "parameter": ["{{ place.name }} ({{ parent.name }})"],
          "required": true
        }
```

inromualdo:
@kennsippell This is a good idea. We could display a text/label in the formview

PR #108, opened 2024-03-19
Autogenerate VHT Area Name, CHP Area Name, and VHT's Village 

kennsippell:
#47 

A few notes about trying to get rid of [this line in gross.ts](https://github.com/medic/cht-user-management/pull/108/files#diff-93a51a116fae565ca730cdce509d40f1dc8cc5b7a111219cbb1892826103929bR13):

I attempted to implement this with config interface like `{ "type": "name", "template": "{{ contact.name }} Area" }` so we can do something like prompt for a CHU name AND consistently append something to the end. I added `{{ this }}` to the templates. This worked fine and I really liked how it separated "generation code" from "validation code" - our validation library is poorly name as it is far from just validation.

This created some place resolution issues which I didn't debug, but also resulted in a new problem where the username was `CHU_community_health_unit`. I didn't like that and also didn't know how to resolve it, so I backed off and went with the approach you see in this PR.


src/config/chis-ke/config.json
```diff
@@ -266,7 +267,7 @@
           "friendly_name": "CHA Name",
           "property_name": "name",
           "type": "name",
-          "parameter": ["\\(PHO\\)", "\\(CHA\\)"],
+          "parameter": ["\\(\\s*PHO\\s*\\)", "\\(\\s*CHA\\s*\\)"],
```
kennsippell, line 270: note this change was just during testing. i saw inputs like Richard Nixon (PHO ) which were not being properly formatted

src/routes/files.ts
```diff
@@ -35,9 +25,9 @@ export default async function files(fastify: FastifyInstance) {
         place.name,
         place.contact.properties.name,
         place.contact.properties.phone,
+        place.userRoles.join(' '),
```
kennsippell, line 28: just reordering this. i personally prefer having username/password at the very end for easy access

README.md
```diff
@@ -61,14 +61,50 @@ required | boolean | True if the object should not exist without this informatio
...
+| phone     | A valid phone number for the specified locality        | Auto formatting provided by [libphonenumber](https://github.com/google/libphonenumber)          | Two letter country code specifying the locality of phone number (eg. `"parameter": "KE"`)             |
+| none      | None                                                   | None                                                                | None                                                                                                   |
+| gender    | A binary gender (eg. `Male`, `Woman`, `M`)             | Formats to either `Male` or `Female`                                | None                                                                                                   |
+| generated | None. No user inputs.                                  | Uses [LiquidJS](https://liquidjs.com) templates to generate data    | None                                                                                                   | [Details](#TheGeneratedConfigPropertyType)
```
paulpascal, line 72: not really important but for the anchor link to work it needs to be something like `[Details](#The-Generated-ConfigPropertyType)`

paulpascal (approved):
All right @kennsippell

Issue #61, opened 2024-02-12
Summary of failures after upload


kennsippell:
Users shouldn't need to search for failures afterward.  Could the progress bar also show errors?  Or allow users to filter to show rows which failed/unvalidated/warning?

PR #111, opened 2024-03-20
Implement filters in the directive area

kennsippell:
#61 

image

* Click on the tags in the directive to apply a filter
* Click the `x` beside the active filter to remove it
* User gets a warning `4 results are hidden by a filter`
* Coordinates color and words in tags in both directive and the list of staged users


alexosugo (approved):
Tested filters and progress bar. Working as expected. Good stuff @kennsippell

PR #96, opened 2024-03-14
Add select_one and select_multiple types

inromualdo:
fix #55

PR #122, opened 2024-03-26
add age widget

### LINKED ISSUE #59: DoB input should support "numbers of years"
it should support both age as a number eg 37 and DoB in the `yyyy-mm-dd` format (at least in the generated json)
for age once someone submits a number (37) we can use today's date to calculate dob and save that dob in the json
support age because some people have no idea when their dob is but can tell their age

_Originally posted by @jonathanbataire in https://github.com/medic/cht-user-management/issues/41#issuecomment-1918429716_

freddieptf:
This PR adds an age input that only supports integers. It doesn't completely fix #59 and only provides a way to config the app to accept either one of Calendar or Age integer input @jonathanbataire and convert that to a date of birth at upload. I couldn't find a clean way to hack this one in as it would require having a switch that can dynamically update an input's type. Did you have any ideas here @inromualdo?


freddieptf:
I've updated the PR to add support for a plain date string, the calendar widget will not be shown on the UI. Does this work for you @jonathanbataire?

freddieptf:
While doing this i felt like we might benefit from introducing new property types e.g an "OR" type that can dynamically switch between two input types and correctly handle this in the presentation, validation and data layer. I'm not sure how much of a rewrite we'd need to do this but it might be worth it if we want dynamic inputs on the UI at least.

kennsippell (changes requested):
Good start thanks!

Why can't we just have UI for the `dob` type which allows both inputs? 
I also proposed that we only handle year inputs when coming from CSV and we not handle them in the UI... I never heard feedback on that; but personally I think that's probably what we should do.

freddieptf:
> Why can't we just have UI for the dob type which allows both inputs?

>     the calendar widget will not be shown on the UI
> 
> Is this change pending? I personally think we should leave UI as-is. Or what is your reason to remove the calendar widget?

Because the dob type opens the browser's calendar widget, it's not a normal text input so we can't support both.

test/lib/validation.spec.ts
```diff
@@ -54,6 +54,14 @@ const scenarios: Scenario[] = [
   { type: 'dob', prop: '2016-05-25', isValid: true, altered: '2016-05-25' },
   { type: 'dob', prop: ' 20 16- 05- 25 ', isValid: true, altered: '2016-05-25' },
 
+  { type: 'age', prop: '', isValid: false, altered: '' },
+  { type: 'age', prop: '20', isValid: true, altered: '20 years' },
+  { type: 'age', prop: ' 20 ', isValid: true, altered: '20 years'  },
+  { type: 'age', prop: 'abc', isValid: false, altered: 'abc' },
+  { type: 'age', prop: '  1 0   0 ', isValid: true, altered: '100 years' },
+  { type: 'age', prop: '2000-01-01', isValid: true, altered: '2000-01-01' },
```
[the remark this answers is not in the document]
freddieptf: `age` is converted to `dater_of_birth` when creating the payload

test/lib/validation.spec.ts
```diff
@@ -54,6 +54,14 @@ const scenarios: Scenario[] = [
   { type: 'dob', prop: '2016-05-25', isValid: true, altered: '2016-05-25' },
   { type: 'dob', prop: ' 20 16- 05- 25 ', isValid: true, altered: '2016-05-25' },
 
+  { type: 'age', prop: '', isValid: false, altered: '' },
+  { type: 'age', prop: '20', isValid: true, altered: '20 years' },
+  { type: 'age', prop: ' 20 ', isValid: true, altered: '20 years'  },
+  { type: 'age', prop: 'abc', isValid: false, altered: 'abc' },
+  { type: 'age', prop: '  1 0   0 ', isValid: true, altered: '100 years' },
```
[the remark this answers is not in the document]
freddieptf: I am reusing the existing dob validator for when the input is a date string

src/services/place.ts
```diff
@@ -155,6 +157,26 @@ export default class Place {
...
 
+    const buildContactProperties = (properties: any) => {
+      const props = filteredProperties(properties);
+      return Object.keys(props).reduce((acc: any, key: string) => {
+        if (key === 'age') {
+          const dobValidator = new ValidatorDateOfBirth();
+          if (dobValidator.isValid(props[key])) {
+            acc.date_of_birth = props[key];
+          } else {
+            const age: string[] = props[key].split(' ');
```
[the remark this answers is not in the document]
freddieptf: I opted to preserve whatever the user inputs and only do the conversion at upload time if necessary. Putting it in the validators format/alter function would mean the user would input a number for age and then find a date string on UI which could cause confusion

src/config/chis-ug/config.json
```diff
@@ -99,9 +99,9 @@
           "required": true
         },
         {
-          "friendly_name": "Date of Birth",
-          "property_name": "date_of_birth",
-          "type": "dob",
+          "friendly_name": "Age",
+          "property_name": "age",
```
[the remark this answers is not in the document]
freddieptf: The config is changed but the data schema isn't, this field will be converted to `date_of_birth` at upload time

kennsippell:
> so we can't support both

but you could add like a checkbox or something like we do in cht?

test/lib/validation.spec.ts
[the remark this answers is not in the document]
kennsippell: the `age` validator supports both dob and numeric input...  why would the user ever want two different validators that capture the same type of input? my opinion is there should only be one "dob" or "age" validator that handles a robust set of inputs

src/services/place.ts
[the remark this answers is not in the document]
kennsippell: i dont think that's confusing

src/config/chis-ug/config.json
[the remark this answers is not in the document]
kennsippell: because you hardcode `date_of_birth` in place.ts? is that a good design?

freddieptf:
> I couldn't find a clean way to hack this one in as it would require having a switch that can dynamically update an input's type

It comes back to this, I've pushed changes to effect the changes requested for further discussions

test/lib/validation.spec.ts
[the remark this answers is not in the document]
freddieptf: Not sure what you mean, there's no way to know if the age is an integer or a date without calling the validators

test/lib/validation.spec.ts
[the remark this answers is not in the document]
freddieptf: I added the age input as a separate input that supports both ints/date, it wasn't meant to replace the existing dob input. If you feel like we no longer need a standalone dob input, i can remove it.

src/config/chis-ug/config.json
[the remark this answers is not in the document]
freddieptf: None of the contact fields really change in cht releases but I've reverted this

test/lib/validation.spec.ts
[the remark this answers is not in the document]
freddieptf: Reverted to using to dob

src/lib/validation.ts
```diff
@@ -30,7 +30,7 @@ type ValidatorMap = {
 };
 
 const TypeValidatorMap: ValidatorMap = {
-  dob: new ValidatorDateOfBirth(),
+  dob: new ValidatorAge(),
```
kennsippell, line 33: here we have a `ValidatorAge` class exported from a file named `validator-dob` for a type called `dob`. Can we be consistent? Call everything Age or everything DoB?

src/lib/validator-dob.ts
```diff
@@ -30,3 +30,31 @@ const parse = (input: string) => {
   const strippedInput = input.replace(/ /ig, '');
   return DateTime.fromISO(strippedInput);
 };
+
+export class ValidatorAge implements IValidator {
```
kennsippell, line 34:
```suggestion
export default class ValidatorAge implements IValidator {
```

src/lib/validator-dob.ts
```diff
@@ -30,3 +30,31 @@ const parse = (input: string) => {
...
+  private dobValidator = new ValidatorDateOfBirth();
+
+  private trimSpace(input: string): string {
+    return input.replace(/\s/g, '');
+  }
+
+  isValid(input: string): boolean {
+    const age = Number(this.trimSpace(input));
+    return (!isNaN(age) && age > 18 && age < 200) || (this.dobValidator.isValid(input));
```
kennsippell, line 43: can you make a 10y/o user with a date input but not an int? at minimum we should inforce this for all inputs.

why `18`? dob has 0 hours.... where is this requirement from?  i think doubt cht-user-management should be the thing that stops an 18yo CHW from entering the eCHIS system....  should it be 10 or something which validates sanity but isn't too controversial?

src/routes/add-place.ts
```diff
@@ -31,6 +31,19 @@ export default async function addPlace(fastify: FastifyInstance) {
     return resp.view('src/liquid/app/view.html', tmplData);
   });
 
+  fastify.post('/widget', async (req, resp) => {
+    const queryParams: any = req.query;
+    const contactType = Config.getContactType(queryParams.place_type).contact_properties.find(prop => prop.type === queryParams.input);
```
kennsippell, line 36: is `input` the right name for this parameter?

src/routes/add-place.ts
```diff
@@ -31,6 +31,19 @@ export default async function addPlace(fastify: FastifyInstance) {
...
+    return resp.view('src/liquid/components/contact_type_property.html', {
+      data: req.body,
+      include: {
+        prefix: 'contact_',
+        place_type: queryParams.place_type,
```
kennsippell, line 41: `const { place_type, input } = queryParam;`

This quick alias can be useful to increase code readability - reducing line length, making refactoring easier, etc

test/lib/validation.spec.ts
```diff
@@ -53,7 +54,11 @@ const scenarios: Scenario[] = [
   { type: 'dob', prop: '2030-05-25', isValid: false },
   { type: 'dob', prop: '2016-05-25', isValid: true, altered: '2016-05-25' },
   { type: 'dob', prop: ' 20 16- 05- 25 ', isValid: true, altered: '2016-05-25' },
-
+  { type: 'dob', prop: '20', isValid: true, altered: DateTime.now().minus({ years: 20 }).toISODate() },
```
kennsippell, line 57: should the tests mock now and hardcode the expected output?

src/liquid/components/contact_type_property.html
```diff
@@ -1,17 +1,39 @@
-{% if include.prop.type != 'generated' %}
+<div id="div_{{include.prop.property_name}}">
+  {% if include.prop.type != 'generated' %}
   {% capture prop_name %}{{ include.prefix }}{{include.prop.property_name}}{% endcapture %}
+  {% if include.prop.type == 'dob' %}
```
kennsippell, line 4: can we make a `contact_type_dob.html` (or similar)?  i suspect that'd be cleaner and can hopefully be used directly by your api endpoint

src/liquid/place/create_form.html
```diff
@@ -25,7 +25,7 @@
   <section class="section is-small">
     {% for prop in contactType.contact_properties %}
       {%
-        include "components/contact_type_property.html" prefix="contact_" prop=prop
+        include "components/contact_type_property.html" prefix="contact_" prop=prop place_type=contactType.name
```
kennsippell, line 28: is this required on line21 also?

src/routes/add-place.ts
```diff
@@ -31,6 +31,19 @@ export default async function addPlace(fastify: FastifyInstance) {
     return resp.view('src/liquid/app/view.html', tmplData);
   });
 
+  fastify.post('/widget', async (req, resp) => {
+    const queryParams: any = req.query;
+    const contactType = Config.getContactType(queryParams.place_type).contact_properties.find(prop => prop.type === queryParams.input);
+    return resp.view('src/liquid/components/contact_type_property.html', {
+      data: req.body,
+      include: {
+        prefix: 'contact_',
```
kennsippell, line 40: can a place not have a dob? how is that controlled?

test/lib/validation.spec.ts
```diff
@@ -53,7 +54,11 @@ const scenarios: Scenario[] = [
...
+  { type: 'dob', prop: ' 20 ', isValid: true, altered: DateTime.now().minus({ years: 20 }).toISODate() },
+  { type: 'dob', prop: 'abc', isValid: false, altered: 'abc' },
```
kennsippell, line 59: can you add a test under your minimum age? over your maximum?

src/routes/add-place.ts
```diff
@@ -31,6 +31,19 @@ export default async function addPlace(fastify: FastifyInstance) {
     return resp.view('src/liquid/app/view.html', tmplData);
   });
 
+  fastify.post('/widget', async (req, resp) => {
```
kennsippell, line 34: is widget a descriptive name for what this api endpoint is doing? i don't think we use that language anywhere else ... what is a widget?

this is for dob only, right? i think the name should likely include dob..
don't all routes in this file start with `/place`? should this also?

kennsippell (changes requested):
Great stuff

src/liquid/components/contact_type_property.html
kennsippell: especially with this inbound also https://github.com/medic/cht-user-management/pull/96/files#diff-cffbfbd30f96baa3bf1eecaf18f63d0fce51c820e87c0cf589510bd88bb2ce1eR8

src/routes/add-place.ts
freddieptf: changed to property_type

src/lib/validator-dob.ts
freddieptf: I've enforced that dates should be less than the current now. I used 18 on the assumption that anyone at or below is a minor. Until there's an explicit request from a partner for the minimum then i think this should be enough

src/routes/add-place.ts
freddieptf: I guess technically it can. I've decided to use whatever prefix is currently being used on the form

test/lib/validation.spec.ts
freddieptf: I don't see the value if mocking `now` here, this mirrors the code that will run in production more. Unless we're thinking of changing the date time lib?

src/routes/add-place.ts
freddieptf: > is widget a descriptive name for what this api endpoint is doing? i don't think we use that language anywhere else ... what is a widget?

oh I just left it in as i was experimenting with how we'd support this kind of inputs in the future. I'm not really sure where this endpoint should live but I've renamed it to `/place/dob`.

src/lib/validator-dob.ts
```diff
@@ -31,21 +31,25 @@ const parse = (input: string) => {
   return DateTime.fromISO(strippedInput);
 };
 
-export class ValidatorAge implements IValidator {
+export default class ValidatorAge implements IValidator {
```
kennsippell, line 34: why is class ValidatorAge exported from the file validator-dob?

src/routes/add-place.ts
kennsippell: it is only used by the app-place form, so having it in here seems "reasonble"
the routes aren't super clearly designated. if you ever want to clean it up i welcome it!

src/lib/validation.ts
```diff
@@ -4,7 +4,7 @@ import Place from '../services/place';
 import RemotePlaceResolver from './remote-place-resolver';
 import { RemotePlace } from './cht-api';
 
-import ValidatorDateOfBirth from './validator-dob';
+import ValidatorDob from './validator-dob';
```
kennsippell, line 7:
```suggestion
import ValidatorDoB from './validator-dob';
```

or DateOfBirth? Dunno how common that acronym is...?

src/lib/validator-dob.ts
```diff
@@ -30,3 +30,35 @@ const parse = (input: string) => {
...
+      return DateTime.now().minus({years: age}).toISODate();
+    }
+    if (this.dobValidator.isValid(input)) {
+      return this.dobValidator.format(input);
+    }
+    return input;
+  }
+
+  get defaultError(): string {
+    return 'Age should be a number between 18 and 200 or a valid date (eg. 1990-02-26)';
```
kennsippell, line 62: no longer has 18/200 limits

src/liquid/components/contact_type_property.html
kennsippell: don't agree with this? no worries if not.

src/routes/add-place.ts
```diff
@@ -31,6 +31,19 @@ export default async function addPlace(fastify: FastifyInstance) {
...
   });
 
+  fastify.post('/place/dob', async (req, resp) => {
+    const { place_type, prefix, prop_type } = req.query as any;
+    const contactType = Config.getContactType(place_type).contact_properties.find(prop => prop.type === prop_type);
+    return resp.view('src/liquid/components/contact_type_property.html', {
+      data: req.body,
+      include: {
+        prefix: prefix,
+        place_type: place_type,
```
kennsippell, line 41:
```suggestion
        prefix,
        place_type,
```

src/liquid/components/contact_type_property.html
```diff
@@ -1,17 +1,39 @@
...
   <div class="field">
     <label class="label">
+      {% if data['option_dob_unknown'] %}
+      Age (Years)
```
kennsippell, line 8:
```suggestion
      Age in Years
```

?

kennsippell (approved):
Smooth

kennsippell:
Completing this PR for Fred since he is on holiday for a week or so.

PR #123, opened 2024-03-28
Remove certificate-arn reference from deploy

nydr:


nydr:
See https://github.com/medic/cht-user-management/pull/112#discussion_r1539531711 for more details

mrjones-plip (approved):

mrjones-plip:
@nydr - there's some [more instances](https://github.com/search?q=org%3Amedic+2023-wildcard-app&type=code) of that `2023-wildcard-app` certificate-arn out there, we're removing those because we want the charts to fall back to the default one, yeah? Should we chase these down and remove them too?

kennsippell (approved):

Hareet:
> @nydr - there's some [more instances](https://github.com/search?q=org%3Amedic+2023-wildcard-app&type=code) of that 2023-wildcard-app certificate-arn out there, we're removing those because we want the charts to fall back to the default one, yeah? Should we chase these down and remove them too?

Those were changed directly on k8s. Bad practice :(

We don't have a build server and helm isn't our main source of truth, so I hot-fixed :(

kennsippell:
Deployed this in prod, working great

Issue #124, opened 2024-03-29
Replacing a CHP results in ambiguous match for every CHP in the area


kennsippell:
CSV input

```csv
Outgoing CHP,CHU,Sub County,CHP Name,CHP Phone
dne,Sawawa,Malava,Not Florence,0712 344321
```

Import this as CHP to Kakamega instance

Expected - DNE CHP does not exist
Observed - Every CHP area in Sawawa is a match and error shows ambiguous match

kennsippell:
Regression from https://github.com/medic/cht-user-management/pull/108

PR #125, opened 2024-03-29
Fix replacement scenarios when place.name is generated

kennsippell:
#124


paulpascal (approved):
It looks good to me 🍞

Issue #62, opened 2024-02-12
Do we want loading spinners when things are loading? Search boxes even.


kennsippell:

kennsippell:
After #142 we should add one more spinner for when users refresh (both the whole table and the row) since it isn't clear now when the refresh operation is complete.

PR #142, opened 2024-04-17
disable and hide csv import button when request in flight

### LINKED ISSUE #63: Improved performance for large table refresh - Refresh single rows instead of entire table
After staging many users (eg. an entire county), the app exerpience slows down.  Uploads (bulk and single), refreshes, and page load (eg edit) take a long time.  Can we be more selective inwhat we are refreshing to make this experience faster?

### LINKED ISSUE #136: Duplicate requests when button is clicked quickly before state change
Reproducible when clicking `Add` to upload a csv and i believe this should also affect the `Upload` button

freddieptf:
Fixes #136 
Fixes #63


src/server.ts
```diff
@@ -18,9 +19,10 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
   fastify.register(multipart);
   fastify.register(FastifySSEPlugin);
   fastify.register(cookie);
+  fastify.register(fastifyCompress);
   fastify.register(view, {
     engine: {
-      liquid: new Liquid({ extname: '.html', root: 'src/liquid', jekyllInclude: true, dynamicPartials: true }),
+      liquid: new Liquid({ extname: '.html', root: 'src/liquid', cache: true, jekyllInclude: true, dynamicPartials: true }),
```
kennsippell, line 25: looks like it is recommended to enable this only in production environments. can we do that?
https://liquidjs.com/tutorials/use-in-expressjs.html#Caching

i personally like the dev experience of changing the view and having it reload without restarting the service

src/routes/add-place.ts
```diff
@@ -74,10 +74,8 @@ export default async function addPlace(fastify: FastifyInstance) {
           },
         });
       }
-
-      // back to places list
-      resp.header('HX-Redirect', `/`);
-      return;
+      resp.header('HX-Trigger-After-Swap', `GotoList`);
+      return `<div class="control"><progress class="progress is-small is-primary" max="100"></progress></div><span><i>processing</i></span>`;
```
kennsippell, line 78: i don't think we should do this kinda thing. it is too easy to make a change in the progress-bar view and not know you need to update this. 

can we find another way? maybe use something like a view with `hx-swap-oob` instead?

src/liquid/place/bulk_create_form.html
```diff
@@ -4,8 +4,9 @@
     hx-encoding='multipart/form-data' 
     autocomplete="off"
     hx-post="/place?op=bulk&type={{contactType.name}}"
-    hx-target="this" 
-    hx-swap="outerHTML"
+    hx-disabled-elt="button#place_create_submit"
```
kennsippell, line 7: i like the use of `hx-disable-alt` why do we need to do all the other stuff? are you trying to hide it?
is disabling the button not enough?

src/liquid/place/bulk_create_form.html
```diff
@@ -110,6 +111,10 @@
             fileName.textContent = fileInput.files[0].name;
           }
         }
+        const form = document.querySelector("form#form_place_create");
+        form.addEventListener("GotoList", function(e) {
```
kennsippell, line 115: can you explain why you're doing this?

kennsippell (changes requested):
good idea and good change

src/liquid/place/bulk_create_form.html
freddieptf: After uploading the csv there's a period of time, when fetching the next page, that the buttons are still visible. You can still click the `Add` button then. I think i can just use the progress bar here from the start and not use `hx-disabled-elt`

src/liquid/place/bulk_create_form.html
freddieptf: I am doing this so i can swap the `Add` view with the progress bar before initiating the redirect to the product list. For this i trigger a custom event that signals when we're ready to redirect. Before, the redirect happened immediately but since the place list takes some time to load, the bulk upload form will still be visible until the request for the the place list page completes

src/liquid/place/bulk_create_form.html
kennsippell: Is `hx-indicator` useful?
Could we maybe use an animated spinner (https://github.com/medic/cht-user-management/issues/62) instead making them empty?

src/liquid/place/bulk_create_form.html
kennsippell: Wonder if chunking the table response would be another option?  So rather than waiting for the whole table to be ready and the whole response to come in, we could send some minimal chunk.

In my mind, this might be a little better because it avoids the complexity of this custom event and the swapping complexity + it is a good practice for large pages

src/liquid/place/bulk_create_form.html
freddieptf: Maybe not, the indicator will be visible for the duration of the csv upload request. When that requests completes there's still the transition period between the upload form and the next page where a user could still click the `Add` button

src/liquid/place/bulk_create_form.html
freddieptf: Sounds good, we can definitely show the table then lazy load the data

freddieptf: I've implemented the change, first we load the skeleton table then load the items.

src/liquid/place/bulk_create_form.html
kennsippell: Is this still a problem now that we lazy-load the data?

freddieptf: I also think It's safe to ignore now, the user would need a really bad connection or sub-100ms reaction time to view/click that button

kennsippell:
The performance tweaks in this PR are spot-on.

For generation times, measured the change to enable `cache: true` in LiquidJS and it brought down time-to-generate by 80% (2.5s down to 0.5s).  For response size, it brought responses sizes down 97% (from 3.5MB to ~125kb).

src/server.ts
```diff
@@ -18,9 +19,16 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
...
   fastify.register(FastifySSEPlugin);
   fastify.register(cookie);
+  fastify.register(fastifyCompress);
   fastify.register(view, {
     engine: {
-      liquid: new Liquid({ extname: '.html', root: 'src/liquid', jekyllInclude: true, dynamicPartials: true }),
+      liquid: new Liquid({ 
+        extname: '.html', 
+        root: 'src/liquid', 
+        cache: process.env.NODE_ENV === 'production',
```
kennsippell, line 28: The performance tweaks in this PR are spot-on.

For generation times, measured the change to enable `cache: true` in LiquidJS and it brought down time-to-generate by 80% (2.5s down to 0.5s).  
For response size, compression brought responses sizes down 97% (from 3.5MB to ~125kb).

kennsippell (changes requested):
Sorry! I didn't send the code review... I left in pending state

kennsippell (changes requested):
I didn't see this review was pending. Apologies for the slowness.

* I just did a quick test and I think the buttons in the rows aren't working - try clicking "upload", "trash can", or "refresh" on any row in the table and nothing happens.
* During CSV upload could we maybe disable the cancel button too? It's clickable but I'm not sure if it does anything.

Code looks good!

freddieptf:
> I just did a quick test and I think the buttons in the rows aren't working - try clicking "upload", "trash can", or "refresh" on any row in the table and nothing happens.

oh i missed that, i think it's a combination of how swaps are done on the table and https://github.com/bigskysoftware/htmx/pull/2247. I've updated the libraries and added support for individual row updates

freddieptf:
@kennsippell I've made some changes to the events api and the updates should appear "faster" now. This should fix https://github.com/medic/cht-user-management/issues/63 as well

src/routes/app.ts
```diff
@@ -60,22 +77,21 @@ export default async function sessionCache(fastify: FastifyInstance) {
...
   // initiates place creation via the job manager
-  fastify.post('/app/apply-changes', async (req) => {
+  fastify.post('/app/apply-changes', async (req, resp) => {
     const uploadManager: UploadManager = fastify.uploadManager;
     const sessionCache: SessionCache = req.sessionCache;
 
     const chtApi = new ChtApi(req.chtSession);
     uploadManager.doUpload(sessionCache.getPlaces(), chtApi);
+
+    return resp.view('src/liquid/place/directive_3_in_progress.html');
```
kennsippell, line 91: shouldn't this just be the directive.html view and let it route accordingly?

src/routes/events.ts
```diff
@@ -4,43 +4,57 @@ import { Config } from '../config';
...
+        resp.sse({ event: `update-${arg}`, data: '<tr></tr>' });
+      } else {
+        fastify
+          .view('src/liquid/components/place_item.html', {
+            contactType: {
+              ...place.type,
+              hierarchy: Config.getHierarchyWithReplacement(place.type, 'desc'),
+              userRoleProperty: Config.getUserRoleConfig(place.type),
+            },
+            place: place,
```
kennsippell, line 40:
```suggestion
            place,
```

kennsippell (approved):
Excellent change!
Remember to update package version before merge.

kennsippell (changes requested):
I've got a problem when submitting invalid csv files. No error is displayed.

src/services/upload-manager.ts
```diff
@@ -82,24 +82,14 @@ export class UploadManager extends EventEmitter {
...
+  public triggerRefresh(place_id: string) {
+    this.emit('refresh_table_row', place_id);
   }
 
   private eventedPlaceStateChange = (subject: Place | Place[], state: PlaceUploadState) => {
     if (!Array.isArray(subject)) {
       subject = [subject];
     }
-    
-    if (subject.length > 1) {
```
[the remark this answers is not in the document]
freddieptf: > when you edit CHP A you can cause changes to CHP B. I suppose I will now have to monitor all changes on all rows and separately trigger each refresh

Triggering individual events is the most optimal. Refreshing the table is basically just refreshing the page at this point

src/liquid/components/place_item.html
```diff
@@ -64,8 +63,7 @@
 
     <td>
         {% if place.creationDetails.password %}
-        {% capture explanation %}Username: {{ place.creationDetails.username }}
-        Password: {{ place.creationDetails.password }}{% endcapture %}
+        {% capture explanation %}Username: {{ place.creationDetails.username }} Password: {{ place.creationDetails.password }}{% endcapture %}
```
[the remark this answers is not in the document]
freddieptf: Anything after the new line will be ignored when pushing the html via sse https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format

kennsippell (approved):

PR #148, opened 2024-04-18
Add certificate-arn for 2024

nydr:
Turns out this is needed after all and should be 2024, fyi @Hareet 
I've rolled this out manually on ingresses since the versions differs from the values files here
- Related #123


mrjones-plip (approved):
looks good - and nice job moving `2023` -> `2024`

nydr:
FYI @kennsippell

Issue #150, opened 2024-04-21
user manager accounts on some instances experiencing 403s - created without mm-online role


kennsippell:
If you POST to `/api/v2/users` and create a user with an online role, then it will automatically be assigned `mm-online` role.
Users on most instances (eg. Busia) have the mm-online role and are working fine.
Users created on the kajiado instance do not have mm-online role and are therefore experiencing 403 errors for routine tasks.

The most likely cause here was that the eCHIS config with the `user_manager` role wasn't uploaded to the instance at the time the users were created.

kennsippell:
Also impacting tharakanithi.  Blocking from checking other instances

kennsippell:
Fixed for Kajiado and Tharaka Nithi.  Leaving this open because I'm blocked fixing the *-echis.health instances.

kennsippell:
Fixed another 25 users on training-1 instance

kennsippell:
Another batch fixed on chis-staging.health.go.ke

PR #151, opened 2024-04-22
Explicitly set mm-online role when creating user managers

kennsippell:
This is a mitigation to prevent #150 from happening again in the future


paulpascal (approved):
Good to me ✨

PR #157, opened 2024-04-25
Improve the UI for login

### LINKED ISSUE #117: Design Changes - Login Page
We need to visually align the page ie the logo and title
We can change some of the colors used in the application to allow for accessibility. Eg the green color used for the sign in button. we can possibly look at using CHT colors.

Wireframe of a proposed redesign

image

From #67

### LINKED ISSUE #131: Design Changes - Directive changes
- [ ] The "getting started" directive should become a centered, prominent "getting started" page
- [ ] If there is no summary to display, remove it until users are added
- [ ] Add spacing between buttons for a cleaner look
- [ ] Summary should likely be horizontal outside of the directive

image

- [ ] The "upload when ready" and "download resulting credentials" can be "notifications" so they both appear when needed

image

- [ ] Show login information under cloud dropdown

jonathanbataire:
## Description
This PR is 1/3 of #172 and #168
Fixes #131 
Fixes #117 
### Changes 

- dropped "tool" in the system name so "CHT User Management"
- rename "sign in" to "login"


src/liquid/auth/authentication_form.html
```diff
@@ -1,39 +1,32 @@
...
+        <div><input name="username" type="text" placeholder="Username"></div>
+        <div><input name="password" type="password" placeholder="Password"></div>
       </div>
     </div>
-    <div class="field">
-        <div class="control">
-            <input name="password" class="input" type="password" placeholder="password">
-        </div>
+    {% if errors %}
+    <div class="error center">
```
kennsippell, line 22: Can we use [hx-disable](https://htmx.org/attributes/hx-disable/) to disable all inputs and the login button to prevent multiple clicks?

src/liquid/auth/authentication_form.html
```diff
@@ -1,39 +1,32 @@
...
     </div>
-    <div class="field is-grouped is-grouped-right">
-      <p class="control">
-        <button class="button is-primary">
-          Sign In
-        </button>
-      </p>
+    {% endif %}
+    <div class = "">
+      <button class=" login-btn">
```
kennsippell, line 27: When I click `login`, it sometimes takes like 4-5 seconds before I see an error.
When I retry the login, the error remains and there is no indication that anything is happen. I find this state a bit difficult - waiting, with an error visible, with no indication anything is happening.
Can we do something?  Maybe hx-disable is enough (feedback above). Maybe we could do something with [hx-indicator](https://htmx.org/attributes/hx-indicator/) like in https://github.com/medic/cht-user-management/pull/142?

kennsippell (changes requested):
Nice work Jonathan

src/public/css/styles.css
```diff
@@ -0,0 +1,95 @@
+/* app colors */
+/* font sizes */
+.theme, html,
+body {
+  background-color: #323232;
```
kennsippell, line 5: Is it confusing to copy the look and feel of CHT? Or is that what we want to do here? 

I'm concerned that user managers sometimes have to use the cht and sometimes they have to use this management tool; and if they look too similar this might be confusing for some of them.  

Maybe we shouldn't match one prominent thing like the background color?  I also think this color is difficult since we have text in the logos.  You're using a white font on this background; but the logo for Kenya has black font.  We can't use a white font in the logo because it is also used on the other pages which have a white background.

src/public/css/styles.css
```diff
@@ -0,0 +1,96 @@
+/* app colors */
+/* font sizes */
```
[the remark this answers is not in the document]
jonathanbataire: - yes this is compiled from ./src/scss/styles.scss should we move this to a dist folder to eliminate confusion?
- Yes tried to do this with custom scss i fill like bulma makes the html look bloated
with this we compile only whats required
makes it look not clean :) 
overall i think html looks cleaner with this approach 
do you have strong opinions on bulma? i can try it again there  lmk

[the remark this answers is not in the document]
jonathanbataire: iwhen it comes to maintainability yes it can get complicated 
but be trying to make it understandable and organized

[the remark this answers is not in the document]
jonathanbataire: gonna try it in with bulma!

src/liquid/auth/authentication_form.html
jonathanbataire: cant seem to get this to work sos

src/public/css/styles.css
```diff
@@ -0,0 +1,139 @@
...
+  background-color: #F8F8F8;
+  font-family: Noto, sans-serif; }
+
+.btn, .login-btn {
+  border: none;
+  border-radius: 4px;
+  font-size: 1em;
+  cursor: pointer;
+  text-decoration: none;
+  box-shadow: 0px 4px 10px 0px rgba(0, 0, 0, 0.5); }
```
freddieptf, line 18: Can we have the login button on the same elevation as the card it's on? Dropping a shadow from the login button makes it look a level higher than the other elements on the card.

src/liquid/auth/authentication_form.html
```diff
@@ -1,39 +1,48 @@
...
+      <div><input name="username" type="text" placeholder="Username" required></div>
+      <div><input name="password" type="password" placeholder="Password" required></div>
     </div>
   </div>
+
+  <div id="error-msg" class="error center">
+    <span>Login Failed. Please Try Again</span>
+  </div>
+
+  <div id="spinner" class="spinner"></div>
```
freddieptf, line 25: Can we have this spinner show in the login button? I feel like that would be a better indicator of the operation's progress

src/liquid/auth/authentication_form.html
```diff
@@ -1,39 +1,48 @@
...
+
+  <div>
+    <button class="login-btn" hx-disable>Login</button>
```
freddieptf, line 28: You should add this to the form https://htmx.org/attributes/hx-disabled-elt/

src/liquid/auth/authentication_form.html
freddieptf: Take a look at https://v1.htmx.org/attributes/hx-indicator/ and how it's currently used eg in https://github.com/medic/cht-user-management/blob/main/src/liquid/place/bulk_create_form.html

src/liquid/auth/authentication_form.html
```diff
@@ -1,39 +1,48 @@
...
+  <div>
+    <label>Credentials</label>
+    <div>
+      <div><input name="username" type="text" placeholder="Username" required></div>
+      <div><input name="password" type="password" placeholder="Password" required></div>
     </div>
   </div>
+
+  <div id="error-msg" class="error center">
+    <span>Login Failed. Please Try Again</span>
```
freddieptf, line 22: Can we show this below the login button?

freddieptf (changes requested):

src/routes/authentication.ts
```diff
@@ -35,6 +35,8 @@ export default async function authentication(fastify: FastifyInstance) {
     try {
       chtSession = await ChtSession.create(authInfo, username, password);
     } catch (e: any) {
+      // TODO: fine tune error codes to match response
+      resp.status(400);
```
freddieptf, line 39: Why are we setting the status code to 400 and not 401?

jonathanbataire:
@freddieptf this is ready for review

jonathanbataire:
@freddieptf requesting review here this is ready

Issue #165, opened 2024-05-06
New CHP Area shows as invalid until edited


kennsippell:
1. Login to Baringo
2. Set CHU to `Chemususu`, CHP Name `Chp`, phone number `0712344321`

Expected - Valid CHP
Observed - Says invalid

Workaround: click edit, click save

kennsippell:
Regression from https://github.com/medic/cht-user-management/pull/153

PR #166, opened 2024-05-06
Staged places appear invalid when they have required generated properties

kennsippell:
#165


alexosugo (approved):
Tested and verified, no longer appears as invalid.

PR #168, opened 2024-05-11
main UI - directives and table views

### LINKED ISSUE #131: Design Changes - Directive changes

jonathanbataire:
## Description
This PR is  2/3 of #172 and #157
Fixes #131


src/liquid/app/nav.html
jonathanbataire, line 1: renamed to header.html to distinguish between header and nav bar

src/routes/app.ts
jonathanbataire, line 1: apply changes and filter modified to filter by contactType

src/liquid/app/header.html
freddieptf, line 1: I see we removed most of the actions from the nav but i don't see any alternatives introduced for clearing the place list, refreshing the place list, creating places etc

src/liquid/place/list.html
freddieptf, line 1: Can we add an explicit color for the `in_progress` state? It looks like it defaults to gray which i think is a bleak color for a progress state

src/liquid/place/directive.html
freddieptf, line 1: The places don't get filtered when i click on any of the filters

freddieptf (changes requested):

src/scss/_tables.scss
```diff
@@ -0,0 +1,166 @@
+$table-bg-color: #f2f2f2;
+$table-header-bg-color: rgba(#dcdcdc, 0.6);
+
+.table-container {
+    overflow-x: auto;
+    display: flex;
+    margin: 0.5em 5px 1em 5px;
+    width: auto;
+    border-radius: $element-border-radius;
+    @include card-shadow;
```
freddieptf, line 10: I don't think we should have a shadow on the table. In this case, the table and the pagination controls should be on the same level yet the table is "above" the pagination controls. I think we should use shadows sparingly

src/public/scripts/pagination.js
freddieptf, line 1: I feel like the UI would be much snappier if we paginated the Place list from the API. I've noticed this triggering the "page is slow or unresponsive" notification on Firefox on large lists. Currently we're lazy loading the place list but i think a solution around pagination would be much better

src/liquid/app/header.html
jonathanbataire: was following [this](https://docs.google.com/document/d/1hsBrovnn7Nv70D8yZnC6MxatnTEz64grtSFf0YYRj8I/edit) for design thats why options moved away from then header bar

must have missed the clear option though will add that

src/liquid/place/directive.html
jonathanbataire: interesting would like to know how this is tested
what browser are you using? 
this works in my chrome and firefox

jonathanbataire: image

src/public/scripts/pagination.js
jonathanbataire: glad you brought this up 
how big is the test dataset?
also curious on how you are testing this be cause it loads very fast even with large datasets on my end since data is loaded once and this pagination is just ui showing and hiding rows

but the reason i used the UI not the api to paginate was the api is slow
you can test this in prod
- load a large dataset
- filter the dataset, you'll realize the more data you have the slower the filter gets
you can feel performance drop with as little as 30-40 rows
Which brings me to this why are we using session storage to store data? i dont think its the right solution
we could try browser dbs like pouchDB, indexeddb etc they can offer better performance cc @kennsippell

jonathanbataire: but also very curious how its slow on your end could you share video recording of this?
it not expected to be like this at all

jonathanbataire: im exploring using the api atm

freddieptf: > but the reason i used the UI not the api to paginate was the api is slow

Yes, for large lists (1000+) it takes 1-3 seconds to generate the HTML but if we add pagination, we'd only be generating HTML for a smaller dataset making it much faster

> filter the dataset, you'll realize the more data you have the slower the filter gets
> this is actually what breaks my browser during testing but the pagination works fine
> 

Yes that is what is slowing down my browser as well

> Which brings me to this; why are we using session storage to store data? i dont think its the right solution for scaling
> we could try browser dbs like pouchDB, indexeddb etc they can offer better performance

Right now the server acts as the source of truth since this is a server rendered app, there's no to little state on the client side. It might be the wrong one but not for the reasons you stated, memory will always be faster than anything that does File I/O.

src/liquid/place/directive.html
freddieptf: I am on Firefox, eg if i successfully upload say 10 items from that list and click on the `success` filter, the list doesn't get updated and i still see all places regardless of state

src/liquid/app/header.html
freddieptf: Since there's not alternative for the actions in the `Users` drop down menu i think we should just maintain it for now or provide alternatives

src/public/scripts/pagination.js
jonathanbataire: memory is fast for small data but when it comes to large data i think youre better off with a db

jonathanbataire: where is the data stored if i may ask im missing something

jonathanbataire: we could explore redis as well

freddieptf: > memory is fast for small data but when it comes to large data i think youre better off with a db app becomes memory hog and browser hangs

Memory here is RAM on the server or whatever platform this is deployed on. No data is stored on your browser. For this app, if your browser hangs it's either trying to render a massive amount of HTML or some JS is misbehaving. 

> we could explore redis as well

I'd prefer if we used an embedded DB (like sqlite3) to reduce deployment complexity but we can discuss this later

jonathanbataire: until we instruct it to store data locally its ram even in the browser thats why it struggles  :)
anyway ill implement the api pagination and circle back here

jonathanbataire: @freddieptf  I have update pagination to api its works well until pagesize is about 50 rows 
but still pretty much usable 
this ready for review

freddieptf: > its works well until pagesize is about 50 rows

It should be fine for larger lists if you test with `NODE_ENV=production`. Can we now remove the lazy loading and do pagination? I think the lists loads fast enough now.

src/liquid/components/place_item.html
```diff
@@ -1,88 +1,103 @@
...
-        <a hx-post="/place/refresh/{{place.id}}" data-tooltip="Refresh" hx-target="this" hx-swap="none">
-            <span class="material-symbols-outlined">refresh</span>
+        <a class="action-link" hx-post="/place/refresh/{{place.id}}" data-toggle="tooltip" data-placement="bottom"
+        title="Refresh" hx-target="this" hx-swap="none">
+            <img src="public/icons/refresh.svg" alt="refresh">
         </a>
         {% if place.validationErrors == empty %}
-        <a hx-post="/place/upload/{{place.id}}" hx-target="this" hx-swap="none" data-tooltip="Upload">
-            <span class="material-symbols-outlined">upload</span>
+        <a class="action-link" hx-post="/place/upload/{{place.id}}" hx-target="this" hx-swap="none show:window:top"
```
freddieptf, line 91: Not sure if we really want to scroll way from an item we explicitly clicked an action on

freddieptf (changes requested):

src/liquid/app/view.html
```diff
@@ -1,43 +1,34 @@
...
-      {% if view == "list" %}
-        {% include "app/fragment_home.html" %}
-      {% else %}
-        {% include "app/form_switch.html" %}
-      {% endif %}
+  <div class="container">
+    {% include "app/header.html" %}
+    <div class="content">
+      <div id="app">
+        <div id="workflow-content" class="container tile is-parent">
```
jonathanbataire, line 21: clean up

src/liquid/app/view.html
```diff
@@ -1,43 +1,34 @@
...
-  <link rel="stylesheet" href="/public/material-symbols.css" />
-
-  <style>
-    .material-symbols-outlined {
-      margin-right: 5px;
-    };
-  </style>
+  <script src="/public/scripts/htmx.min.js"></script>
+  <script src="/public/scripts/sse.js"></script>
+  <script src="/public/scripts/bulma-calendar.min.js"></script>
```
jonathanbataire, line 8: clean up

src/public/scripts/pagination.js
jonathanbataire: if by lazy loading you mean the spinner logic i beg to differ i think its very useful for bad internet/ heavy server loads that might delay response its good practice we should leave it

src/services/pagination.ts
```diff
@@ -0,0 +1,66 @@
...
+  private page;
+  private pageSize;
+  private cookiePageSize;
+  private cookieCurrentPage;
+  private requestContactTypeName;
+
+  constructor(options: { page: number; pageSize?: number; cookie: {}; requestContactTypeName?: string }) {
+    this.pageSize = options.pageSize ? options.pageSize : 10;
+    this.page = options.page;
+    this.cookiePageSize = this.getPaginationCookie(options.cookie, 'pageSize');
```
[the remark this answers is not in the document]
jonathanbataire: i want to persist page number in cookie to handle all miscellaneous call to root "/" e.g when you move out to create more users/edit etc you should come back to your "current page"
i can create logic to handle the situation you are talking about

src/routes/events.ts
```diff
@@ -30,6 +30,28 @@ export default async function events(fastify: FastifyInstance) {
       });
     };
 
+    const updateNavBadge = async () => {
```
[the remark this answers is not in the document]
jonathanbataire: directive updates the directive, this nav bar is outside the directive

src/services/pagination.ts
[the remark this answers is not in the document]
jonathanbataire: i dont understand is this comment intended for the entire pagination file?  
you mean slice it in front end? that would be expensive for the PC

src/liquid/components/summary.html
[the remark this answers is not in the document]
jonathanbataire: this is the summary essentially place states for each contacttype 
plan is to handle it later in another pr as this is very big already but shouldnt be a blocker 
unless otherwise, i can remove it lmk what you think

src/liquid/place/place_navbar.html
```diff
@@ -0,0 +1,25 @@
...
+        {% include "components/nav_badge.html" %}
+    </a>
+    {% endfor %}
+    <a hx-trigger="click" hx-get="/app/list/summary"
+    {% capture summarytabClass %} {% if directiveModel.activeTab == 'summary' %}active{%endif%}{%endcapture%}
+    id="tab_summary" hx-target="#place_list" hx-swap="outerHTML show:window:top" class="nav-item {{summarytabClass}}" onclick="highlight(this)">Summary</a>
+</div>
+
+<script>
+    function highlight(element) {
```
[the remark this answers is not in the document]
jonathanbataire: Anything ui related that can be done frontend shouldnt do around trip to the server,  find it unnecessary

src/css/custom-bootstrap.css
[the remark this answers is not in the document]
jonathanbataire: i can customized bootstrap to get only the elements i want for this its tooltips, badges and breadcrumbs keeping the file very small
not sure you can do that with bulma

src/liquid/components/place_item.html
jonathanbataire: agreed will revert

src/routes/events.ts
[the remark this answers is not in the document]
jonathanbataire: will try move this inside directive and see

src/services/pagination.ts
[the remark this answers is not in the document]
freddieptf: > i can create logic to handle the situation you are talking about

Yeah but that's just more code. Personally, I like less code. Having the state in the URL for both the current tab and page is much simpler and easier to understand at a glance both on the UI and in code

> i want to persist page number in cookie to handle all miscellaneous call to root "/"

Then maybe we should fix that instead?

src/services/pagination.ts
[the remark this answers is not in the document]
freddieptf: This class as used in the endpoint. You can parse the URL param and slice the place list before returning the response to the browser.

src/routes/events.ts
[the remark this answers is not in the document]
freddieptf: oh i see so this updates the count on the navbar. You can just remove the count from the navbar then, no need to duplicate what's already on the directive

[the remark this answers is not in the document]
freddieptf: This might be trickier than i thought. While the filters work, i find them a bit confusing as they represent the state of both lists. In the screenshot below, all the 4 invalid items are on 1 tab so if i apply the invalid filter when on the other tab I get an empty list and have to manually switch tabs to view the invalid items. 

Screenshot 2024-06-05 at 07-27-38 CHT User Management Tool

I think we should move the filters closer to the table, it would be great if they were directly on the status column. The app only needs to show the user some sort of hint when there are invalid/failed items and these hints can have links that just auto-apply a filter

Screenshot 2024-06-05 at 07-40-37 CHT User Management Tool

src/liquid/components/summary.html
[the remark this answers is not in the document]
freddieptf: If it doesn't have anything on it then there's no need to keep it around

src/liquid/place/place_navbar.html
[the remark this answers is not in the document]
freddieptf: While i agree with you, you still make the trip to the server to fetch the page after clicking the tab. Being a server rendered app, the server manages most of the state in the UI. In this case, you can return both the page and the updated navbar in the response from the server and achieve the same effect without JS

src/liquid/components/table_header.html
```diff
@@ -17,6 +18,16 @@
...
         <th></th>
-        <th></th>
+        <th class="header-actions">
+            <a hx-post="/app/refresh-all" hx-swap="none"
+            data-toggle="tooltip" data-placement="bottom" title="refresh table">
+                <img src="public/icons/refresh-table.svg" alt="refresh">
+            </a>
+            <a hx-post="/app/remove-all/{{contactType.name}}" hx-swap="none" 
+            data-toggle="tooltip" data-placement="bottom" title="delete all"
+            hx-confirm="Are you sure you want to DELETE ALL USERS?">
```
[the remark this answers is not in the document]
jonathanbataire: this is when deleting all users hence the strong wording

src/liquid/place/place_navbar.html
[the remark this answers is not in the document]
jonathanbataire: but htmx, sse etc all run in the frontend and do manipulate ui before even hitting the server!
this is equivalent to asking the server to register/listen for button clicks 
i dont think thats what SSR is

src/liquid/components/table_header.html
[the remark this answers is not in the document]
freddieptf: Before, it was "Clear List" which i think feels less destructive of an action. There's a world where a user could interpret `DELETE ALL USERS` as deleting the users they uploaded to the instance

src/liquid/place/place_navbar.html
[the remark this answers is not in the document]
freddieptf: > but htmx, sse etc all run in the frontend and do manipulate ui before even hitting the server!
this is equivalent to asking the server to register/listen for button clicks

There's nothing fancy going on. We're just issuing AJAX calls to the server on certain events. htmx helps with this and content loading/swapping. We can ditch htmx and just use full page reloads for any state change and the app would work the same as the state is all on the server.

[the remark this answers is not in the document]
jonathanbataire: ditching it would be pointless. you can never get rid of js completely  in web development even if your back end is python
JS was built for web development
The point i was making wasnt about htmx only but all the js served in the public folder all of it runs in the browser 
all the calendar js tooltip js etc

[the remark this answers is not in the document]
freddieptf: Yes i understand some UI elements are complex and need JS to manage their state. To be clear, I'm not saying we should get rid of JS, I'm saying that the application state should be driven by the server as much as possible. Application state in this app should be the same with or without JS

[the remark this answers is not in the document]
jonathanbataire: is this why we have things like this [place.className](https://github.com/medic/cht-user-management/blob/136998e2c79015d08cbb80f9ad4c6fcfab80edf3/src/liquid/components/button_upload.html#L1)? server literally storing and setting classes? you dont find this weird?
UI state and server state should be managed seperately, with only communications between the 2 on the state of  one  another

src/routes/events.ts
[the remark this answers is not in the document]
jonathanbataire: i see no problem showing 0 rows if the filter is applied since as long as we clearly show that there is a filter

src/services/pagination.ts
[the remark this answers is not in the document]
jonathanbataire: whats is much code? :) 
you want us to handle all calls to "/"? that sounds like more code than just handling this situation with a variable and an if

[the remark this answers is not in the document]
jonathanbataire: cookies are good to store and persist user preferences there is nothing sensitive we are storing here that the user cant see on the ui

src/services/pagination.ts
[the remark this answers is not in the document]
jonathanbataire: not sure i understand this well
what you're explaining i believe is what i'm doing here
is only sliced/paginate and sent back to the user
this has no effect on the original places data in store

src/services/pagination.ts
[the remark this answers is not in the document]
freddieptf: > i want to persist page number in cookie to handle all miscellaneous call to root "/"
> 
> you want us to handle all calls to "/"?

Find these calls and replace them with appropriate ones i.e redirecting to the correct page. Calling `/` worked before because there was no pagination

> cookies are good for storing and persisting user preferences besides there is nothing sensitive we are storing here that the user cant see on the ui

The current page and active tab are not user preferences, they are part of the navigation state that should be on the URL

src/services/pagination.ts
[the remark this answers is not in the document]
freddieptf: > Not entirely sure if we need this, managing this state would be much easier if all you had to look at was a URL param 

This is part of the cookie conversation. We don't need cookies for these

src/liquid/place/place_navbar.html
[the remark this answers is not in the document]
freddieptf: All I'm saying is update the tab's state by the same way the contents of the page are updated. We don't need a script to do that.

src/routes/events.ts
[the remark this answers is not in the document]
freddieptf: OK, we can change this behavior later then.

jonathanbataire:
@freddieptf  lmk when you have time we do a peer review online for this PR so i can better understand somethings 
I can be available thursday/friday

src/liquid/place/place_navbar.html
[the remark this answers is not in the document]
jonathanbataire: this has been changed

src/services/pagination.ts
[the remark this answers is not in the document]
jonathanbataire: cookies have been removed

jonathanbataire:
@freddieptf requesting review here

src/liquid/app/header.html
```diff
@@ -0,0 +1,42 @@
...
+                    <li><img src="/public/icons/logout.svg">Logout</li>
+                </a>
+                <li><i>Logged in as {{ session.username }}</i></li>
+            </ul>
+        </li>
+    </ul>
+</nav>
+
+<script>
+    document.addEventListener('click', function () {
```
freddieptf, line 28: Not sure why this listener is added to the document and not the element being targeted

src/liquid/app/header.html
```diff
@@ -0,0 +1,42 @@
...
+                <a href="/logout">
+                    <li><img src="/public/icons/logout.svg">Logout</li>
+                </a>
+                <li><i>Logged in as {{ session.username }}</i></li>
+            </ul>
+        </li>
+    </ul>
+</nav>
+
+<script>
```
freddieptf, line 27: It'd be great to keep scripts like this in an `app.js` or something

src/liquid/app/header.html
freddieptf: aah i see, to hide the drop down menu when we click anywhere on the page, correct?

freddieptf:
Something i noticed when the list had one item. The tool tips get cut off and there's a scrollbar
Screenshot from 2024-07-12 19-07-35

src/liquid/app/header.html
freddieptf: Now I'm confused, since the drop down disappears when it's not hovered on, why is this needed?

jonathanbataire:
stale

Issue #12, opened 2024-01-19
Cloud-Based execution of move-contacts commands


kennsippell:
Currently this tool outputs a `cht` console line which can be executed to move contacts.  Let's take this to the next step and execute that command in the cloud.

kennsippell:
Discussion here to also move households and not just CHP areas https://github.com/moh-kenya/config-echis-2.0/issues/1662

paulpascal:
> Discussion here to also move households and not just CHP areas https://github.com/moh-kenya/config-echis-2.0/issues/1662 + using forms in the CHT to trigger the move (not a UI interface like cht-user-management).

Please @kennsippell, sorry for coming back on this but the issue link you shared could not be found. (404).
This is something the MoH Togo is requesting... and I am really interested in this feature.

thanks

kennsippell:
Notes from design conversation today with @paulpascal.

What states will jobs go through:

1. Pending (waiting)
2. (Not MVP) Flag user for maintenance
3. (Not MVP) Wait for the user to sync
4. (Not MVP) Check sentinel backlog and couch2pg backlog to ensure it is safe
5. Execute the move-contacts + upload-docs
6. Success
7. Fail

What is needed from a Cloud-Based Queue

1. Hostable in docker (rabbitMQ)? Or can it be cloud-based like SQS?
2. Great if had a UI to view status, click to cancel, click to view logs, etc so we don’t have to build that
3. We need to wait for users to sync... so we need to able to pop from queue many times, but decide not to process job for some days

Design questions:

1. What should happen if a user never syncs? Should we eventually move them? What is a reasonable limit?
2. What queuing technology should we use for this?
3. How can a user cancel a job? Learn about the state of the job? Investigate a failed job?

Next Steps:

1. Evaluation of cloud-based queues against requirements above. 
  1. Pick one.
  2. It’d be great to just write this down and kinda get some opinions. Maybe post it on #development or #cht-user-management
2. Pick how things will get pushed into the queue?
  * Options
    * User management tool push into queue
    * UI in CHT 
  * The biggest factor here is WHO should move contacts and reorganize hierarchies? Should they be online users? Who is it in #moh-togo/#moh-uganda/#moh_kenya?
3. Make work thread(s) that pull from the queue and kick of cht-conf move-contacts code (via process, or via import, whatever)

-- End of MVP --

4. Expose the queue’s UI somehow? 
5. Or build a UI to show status, cancel, etc.
[...]

PR #177, opened 2024-05-18
feat(#12): cloud-based execution of move contact 

paulpascal:
Closes: #12 

This PR introduces a cloud-based execution of the `move-contact` feature, enhancing user experience by  replacing the previous `cht-conf` command-line approach.

To achieve this, we've implemented [bullmq](https://docs.bullmq.io/), a powerful and flexible queue system built on Redis. 

Here's a breakdown of the process:

1. Upon initiating a contact move, essential details such as `contact_id`, `parent_id`, `instanceUrl`, and the current `sessionToken` (**encrypted**) are compiled and submitted as a job to the `MOVE_CONTACT_QUEUE`.
   
2. A dedicated worker monitors this queue, processing incoming jobs:
   - It has the flexibility to defer processing based on predefined logic, allowing for scheduled execution.
   - Alternatively, it can proceed to execute the `cht-conf` command via a `child_process`, utilizing all pertinent job data, including the **decrypted** `sessionToken`.
   
3. Upon successful execution, the job is marked as completed. In case of any failures, the job is flagged accordingly. Users can conveniently track job statuses directly from the user management tool's dashboard.

> [!NOTE]
> Enabling `cht-conf` to support session tokens required implementing some modifications, detailed [here](https://github.com/medic/cht-conf/pull/617). 
> (_the session token is initial got from  `_session` request_)


mrjones-plip (changes requested):
Really good work here - great to see a compose file that I suspect works out of the box!

My big question is how we're building and publishing images.  I have some smaller feedback points and tried to make code ` ```suggestion`s  when possible - but lemme know if I got any of those wrong!

paulpascal:
Thanks @mrjones-plip for this great review, I will be going through them one by one and get back to you.
About the publishing, I was also thinking, should it be possible to do it the same way we did for the current version (using publish script). Maybe it will need some updates on the current docker stuff I did? (Forgive if I'm sinning here 🥲 I am not a 😅pro on docker)


cc: @kennsippell

shared/queues/queues.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export const QUEUE_NAMES = {
+  MOVE_CONTACT_QUEUE: 'MOVE_CONTACT_QUEUE',
+  // Lets add more queue names as need
+};
+
+
+export interface IQueueManager {
+  addJob(jobParams: any): Promise<string>;
+  removeJob(queueName: string, jobId: string): Promise<void>;
```
kennsippell, line 13: does anything ever call `removeJob`?  YAGNI

src/lib/move.ts
```diff
@@ -20,13 +24,36 @@ export default class MoveLib {
...
+      fromLineage,
+      jobsBoardUrl
+    };
+  }
+
+  private static getJobName(fromLineage: any, toLineage: any): string {
+    return `move_[${fromLineage[0]?.name}]_from_[${fromLineage[1]?.name}]_to_[${toLineage[1]?.name}]`;
+  }
+
+  private static getJobData(fromId: string, toId: string, chtApi: ChtApi): any {
```
kennsippell, line 49: i don't think JobData should be of type `any`

src/lib/move.ts
```diff
@@ -20,13 +24,36 @@ export default class MoveLib {
...
+    const { ENCRYPTION_KEY } = env;
```
kennsippell, line 50: should you read this from env in `shared/config.ts` along with the other env variables?
can we call it something more descriptive like `QUEUE_PRIVATE_KEY` to differentiate it from `COOKIE_PRIVATE_KEY` which is quite similar in functionality

src/liquid/components/move_result.html
```diff
@@ -9,26 +9,27 @@
   </ul>
   <ul>
     <li><span class="tag is-success" style="margin: 0 10px;">After</span></li>
+    <li><span style="margin: 0 10px;">{{ fromLineage.first.name }}</span></li>
```
kennsippell, line 12: no inline styles please

src/liquid/components/move_result.html
```diff
@@ -9,26 +9,27 @@
...
+      <!-- <button
         class="button is-right"
         style="float: right;"
         onclick="navigator.clipboard.writeText(`{{ command }}`);"
       >
         <span class="material-symbols-outlined">content_copy</span>
       </button>
-      {{ command }}
+      {{ command }} -->
+      A worker is taking care of this job, you can please check it <a target="_blank" style="color:chartreuse; font-weight: bold;" href="{{ jobsBoardUrl }}">here</a>.
```
kennsippell, line 32: what should we actually show here?  one option would be to show nothing and remove this page entirely. then we can just show something like a [toast](https://rfoel.github.io/bulma-toast/)?

docker-compose.yml
```diff
@@ -1,13 +1,47 @@
...
+    build:
+      context: .
+      dockerfile: Dockerfile
     environment:
       - EXTERNAL_PORT=${EXTERNAL_PORT:-3000}
       - COOKIE_PRIVATE_KEY=${COOKIE_PRIVATE_KEY}
       - CONFIG_NAME=${CONFIG_NAME}
+      - REDIS_HOST=redis
+      - REDIS_PORT=6379
+      - MOVE_CONTACT_QUEUE=${MOVE_CONTACT_QUEUE:-MOVE_CONTACT_QUEUE}
```
kennsippell, line 20: i don't think this `MOVE_CONTACT_QUEUE` value is actually loaded from env. it appears to be a hardcoded string. i agree it should not loaded from env

shared/queues/queues.ts
```diff
@@ -0,0 +1,64 @@
+import { v4 } from 'uuid';
+import { Queue } from 'bullmq';
+import { redisConnection } from '../config'; // Adjust the import based on your structure
+
+export const QUEUE_NAMES = {
```
kennsippell, line 5: since there is just one queue. do we need this? YAGNI principle

shared/queues/queues.ts
```diff
@@ -0,0 +1,64 @@
...
+  removeJob(queueName: string, jobId: string): Promise<void>;
+  getQueue(queueName: string): any;
+}
+
+
+export type JobParams = {
+  queueName: string;
+  jobName: string;
+  jobData: any;
+  jobOpts?: any;
```
kennsippell, line 22: what is jobOpts? Should it be `any`?

src/server.ts
```diff
@@ -38,7 +38,7 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
     dir: path.join(__dirname, 'routes'),
   });
   fastify.register(fastifyStatic, {
-    root: path.join(__dirname, '../src/public'),
+    root: path.join(__dirname, '../../src/public'),
```
kennsippell, line 41: what is this change? sorry i'm confused why this is needed now

test/lib/authentication.spec.ts
```diff
@@ -1,7 +1,7 @@
 import chai from 'chai';
 import chaiExclude from 'chai-exclude';
 import { mockChtSession } from '../mocks';
-import Auth from '../../dist/lib/authentication';
+import Auth from '../../dist/src/lib/authentication';
```
kennsippell, line 4: I think this was a bug. I don't think this should be loading from the `dist` folder.

test/lib/move.spec.ts
```diff
@@ -27,11 +28,8 @@ describe('lib/move.ts', () => {
     const contactType = Config.getContactType('c_community_health_unit');
     const sessionCache = new SessionCache();
     
-    const actual = await MoveLib.move(formData, contactType, sessionCache, chtApi());
-    expect(actual.command).to.include('--contacts=chu-id');
-    expect(actual.command).to.include('--parent=to-sub');
-    expect(actual.command).to.include('--url=http://username:password@domain.com', actual.command);
-
+    const actual = await MoveLib.move(formData, contactType, sessionCache, chtApi(), mockQueueManager());
+    expect(actual.jobsBoardUrl).to.exist;
```
kennsippell, line 32: can you add back these assertions? I think these are good tests

test/lib/move.spec.ts
```diff
@@ -18,7 +18,8 @@ describe('lib/move.ts', () => {
     [{ id: 'chu-id', name: 'c-h-u', lineage: ['from-sub'], type: 'remote' }],
   );
 
-  it('move CHU: success', async () => {
+  it('move CHU: success', async () => {    
+    process.env.ENCRYPTION_KEY = 'superimportantkey';
```
kennsippell, line 22: you should write your code so modifying global state isn't required for testing

the way i like to do this, is that `main.ts` or whatever your "entrypoint" is should read in all global state and then that data should be injected downward. this makes it so your `main.ts` can't be unit-tested but everything else can

test/shared/utils.spec.ts
kennsippell, line 1: what is `utils`? I think this no longer exists, right?

tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "worker/**/*", "shared/**/*"],
```
kennsippell, line 2: should we create multiple javascript outputs instead of one which includes both frontend and backend?

worker/config/index.ts
kennsippell, line 1: looks like this file shouldn't be required now that it is in `shared/config/index`

worker/main.ts
```diff
@@ -0,0 +1,39 @@
...
+    } catch (error) {
+      console.error(`Error starting worker for queue "${queueName}":`, error);
+      throw error;
+    }
+  }
+
+  console.log(`🚀 All your workers are listening`);
+}
+
+main().catch((error) => {
```
kennsippell, line 36: is catching this helpful? typically errors are automatically written to console and cause an error state, so this shouldn't be needed and might actually reduce what gets logged (eg. where is the stack?)

worker/main.ts
```diff
@@ -0,0 +1,39 @@
...
+
+config();
+
+interface WorkerConfig {
+  queueName: string;
+  workerClass: new (queueName: string) => any;
+}
+
+// Define the configuration for workers
+const workerConfigs: WorkerConfig[] = [
```
kennsippell, line 14: in general, i like the spirit of what you're doing here to make a "generic worker engine" ... but i'm not sure how much value it is adding.  do you have other workers in mind which is motivating this?

in general, i think it might just be best to keep this simple and have a concrete `move-contacts-worker`. i'd suggest we skip most of this `generic worker engine` stuff

no need for a generic interface `WorkerConfig`. no need to loop over a const array of 1. no need for `jobs/move-contacts` because there is just one.  i like the `worker.ts` abstract stuff which handles some of the generic flow; but I don't get why there are two files for `move-contact-job.process` and `move-contact-job.worker`?

worker/jobs/move-contact/move-contact-job.worker.ts
```diff
@@ -0,0 +1,43 @@
...
+export class MoveContactJobWorker extends BullMQWorker {
+  
+  DELAY_IN_DAYS = 3;
+
+  constructor(queueName: string, processor = new MoveContactJobProcessor()) {
+    super(queueName, processor);
+  }
+
+  async canProcess(job: Job): Promise<boolean> {
+    // TODO: Implement logic here
```
kennsippell, line 23: i think to go-live with this, we need to make a request to the monitoring endpoint (https://nairobi.echis.go.ke/api/v2/monitoring) and confirm that `sentinel.backlog` is less than 1000. 
this wil ensure we don't take down the server by moving contacts too quickly

it should postpone probably 1 hour? something like that?

worker/worker.ts
```diff
@@ -0,0 +1,53 @@
...
+        await this.postpone(job);
+        console.log(`Job skipped (not ready): ${job.id}`);
+        return;
+      }
+
+      const result = await this.processor.process(job.data);
+      if (!result.success) {
+        console.log(`Job ${job.id} failed with message: ${result.message}`);
+        job.log(result.message);
+        throw new Error();
```
kennsippell, line 37: descriptive errors please

worker/worker.ts
```diff
@@ -0,0 +1,53 @@
...
+export abstract class BullMQWorker implements JobWorker {
+  readonly MAX_CONCURRENCY = 1; // Limit concurrency to 1 job at a time
+
+  readonly jobQueueName;
+  private worker: Worker;
+
+  constructor(queueName: string, private processor: JobProcessor<any>) {
+    this.processor = processor;
+    this.jobQueueName = queueName;
+    this.worker = new Worker(this.jobQueueName, async (job: Job) => {
```
kennsippell, line 26: this inline function is doing a lot. can we break it out and not do it inline?

worker/worker.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export interface JobProcessor<T> {
+  process(params: T): Promise<JobResult>;
+}
+
+export type JobResult = { success: boolean; message: string };
+
+
+export abstract class BullMQWorker implements JobWorker {
+  readonly MAX_CONCURRENCY = 1; // Limit concurrency to 1 job at a time
```
kennsippell, line 18: private?

worker/worker.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  readonly jobQueueName;
```
kennsippell, line 20: protected? let's be explicit about scope

worker/worker.ts
```diff
@@ -0,0 +1,53 @@
...
+      }
+
+      const result = await this.processor.process(job.data);
+      if (!result.success) {
+        console.log(`Job ${job.id} failed with message: ${result.message}`);
+        job.log(result.message);
+        throw new Error();
+      }
+
+      await job.updateProgress(100);
```
kennsippell, line 40: what is 100?

worker/worker.ts
```diff
@@ -0,0 +1,53 @@
+import { Worker, Job } from 'bullmq';
+import { redisConnection } from '../shared/config';
+
+export interface JobWorker {
```
kennsippell, line 4: this interface doesn't appear to ever be used.  do we need an abstract class and an interface?

worker/jobs/move-contact/move-contact-job.processor.ts
```diff
@@ -0,0 +1,75 @@
...
+      }
+
+      const decryptedSessionToken = decryptSessionToken(
+        sessionToken, ENCRYPTION_KEY
+      );
+
+      const chtProcess = spawn('cht', [
+        `--url=${instanceUrl}`,
+        `--session-token=${decryptedSessionToken}`,
+        `--force`,
```
kennsippell, line 31: is cht-conf's `upload_docs` folder properly cleaned up between runs?  we need to make sure nothing in that persists

worker/jobs/move-contact/move-contact-job.processor.ts
```diff
@@ -0,0 +1,75 @@
...
+        'move-contacts',
+        'upload-docs',
+        '--',
+        `--contacts=${contactId}`,
+        `--parent=${parentId}`,
+      ]);
+
+      chtProcess.stdout.on('data', (data: any) => {
+        console.log(`Cht command output:\n${data}`);
```
kennsippell, line 40:
```suggestion
        console.log(`cht-conf: ${data}`);
```

worker/worker.ts
```diff
@@ -0,0 +1,53 @@
...
+  canProcess(job: any): Promise<boolean>;
+  postpone(job: any): Promise<boolean>;
+  stop(): Promise<void>;
+}
+
+export interface JobProcessor<T> {
```
kennsippell, line 10: could you just use a function type here ... it's a bit ligheter than creating a class which implements a interface which just has one function?

kennsippell (changes requested):
Amazing work Paul!

tsconfig.json
paulpascal: its better ! Will readjust the structure

worker/main.ts
paulpascal: Yes, thats was kind moving by separating concerns and make it easy to add a new worker in a need.
But you are right since we do not have any other worker in mind. We can keep it simpler and drop the generic stuff.

shared/queues/queues.ts
paulpascal: i missed this one, was thinking were to place it

worker/worker.ts
paulpascal: Sure

paulpascal:
Thanks @kennsippell , great review. Really appreciate. Will be making the right adjustment

tsconfig.json
paulpascal: @kennsippell , I like the idea of having all existing stuff put in frontend.

Just to make sure I get it well, we should be having something like ?

```
src/frontend (all from source moved here)
src/worker (move worker inside src/)
```

src/server.ts
kennsippell: oic. this is a relative path within dist... interesting

kennsippell:
Testing today:

1. If redis service is not running, the worker process doesn't exit

worker/jobs/move-contact/move-contact-job.processor.ts
kennsippell: tested this and it seems cht-conf deletes it at the start of each run with `--force` mode

paulpascal: Yes it deletes  files from the previous execution, but the current will be staying there, I will try to delete right after the execution.

kennsippell: if it is cleaned up, then imo we don't think we need to do anything...  can keep it simple and let cht-conf handle it
in extreme cases like a crash during upload-docs, having these files not cleaned up might aid a recovery... ?  probably not. but maybe.

paulpascal: Sure!

test/lib/move.spec.ts
paulpascal: the  `command` is no more available here, since we don't need anymore?

docker-compose.yml
```diff
@@ -1,13 +1,47 @@
...
+    command: npm run start:worker
+    environment:
+      - NODE_ENV=production
+      - REDIS_HOST=redis
+      - REDIS_PORT=6379
+      - MOVE_CONTACT_QUEUE=${MOVE_CONTACT_QUEUE:-MOVE_CONTACT_QUEUE}
+      - ENCRYPTION_KEY=${ENCRYPTION_KEY}
+      - BOARD_PORT=${BOARD_PORT:-3333}
+    ports:
+      - '${BOARD_PORT-3333}:${BOARD_PORT:-3333}'
```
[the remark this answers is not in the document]
paulpascal: Thanks, finally we will not be exposing any port. It would not be needed for the worker

worker/worker.ts
paulpascal: we probably dont need it. Was testing the progress widget of jobs.

worker/jobs/move-contact/move-contact-job.processor.ts
```diff
@@ -0,0 +1,75 @@
...
+      const decryptedSessionToken = decryptSessionToken(
+        sessionToken, ENCRYPTION_KEY
+      );
+
+      const chtProcess = spawn('cht', [
+        `--url=${instanceUrl}`,
+        `--session-token=${decryptedSessionToken}`,
+        `--force`,
+        'move-contacts',
+        'upload-docs',
```
kennsippell, line 33: can we add `--verbose`

Dockerfile
```diff
@@ -13,6 +13,7 @@ COPY package*.json .
 RUN npm ci --omit=dev
 
 COPY src ./src
```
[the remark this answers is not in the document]
paulpascal: Renaming `src` to `frontend` seems pretty good to me, the nit with that is the number of changes it will cause, like changing the relative path in all files... 😅.

src/liquid/components/move_result.html
```diff
@@ -9,26 +9,28 @@
...
     {% endif %}
     {% endfor %}
-    <li><span style="margin: 0 10px;">{{ fromLineage.last.name }}</span></li>
   </ul>
 </nav>
 
 <section class="hero {% if isError %}is-danger{% else %}is-info{% endif %}">
   <div class="hero-body">
     <div>
-      <button
```
kennsippell, line 24: kill it :D

src/plugins/bullmq.ts
```diff
@@ -0,0 +1,40 @@
+import { FastifyInstance } from 'fastify';
+import fp from 'fastify-plugin';
+
+import { BullMQAdapter } from '@bull-board/api/bullMQAdapter';
+import { createBullBoard } from '@bull-board/api';
+import { FastifyAdapter } from '@bull-board/fastify';
+
+import { MOVE_CONTACT_QUEUE, queueManager } from '../shared/queues';
```
kennsippell, line 8: i don't love that `MOVE_CONTACT_QUEUE` needs to be imported here. could it just be the default for a call to `getQueue()`?

src/liquid/place/move_form.html
```diff
@@ -46,3 +47,29 @@ <h1 class="subtitle">To</h1>
...
+  function handleResponse(event) {
+    console.log('Event', event);
+    if (!event.error) {
+      // Reset form
+      const $form = document.getElementById('form_place_move');
+      $form.reset();
+
+      // Show toast
+      const toastContent = `
+        <p>A worker is taking care of this job, you can please check it
```
kennsippell, line 61:
```suggestion
        <p>This will be moved shortly!  It may take up to 24 hours to process.
```

Something like this? The language is a bit more focused on the user experience instead of the technology.

src/worker/move-contact-worker.ts
```diff
@@ -0,0 +1,160 @@
...
+  contactId: string;
+  sessionToken: string;
+  instanceUrl: string;
+}
+
+export type JobResult = { success: boolean; message: string };
+
+export class MoveContactWorker {
+  private readonly DELAY_IN_MILLIS = 360_000;   // 1 hour
+  private readonly MAX_TIMEOUT = 180_000;       // 30 minutes timeout
```
kennsippell, line 20: These are going to be very awkward to try to recover if large moves ever crash or are aborted.  Can we make this longer - like 60 or 90 minutes? For now at least, I don't think we want to be terminating jobs as they process.

test/worker/move-contact-worker.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+import { queueManager } from '../../src/shared/queues';
+import { 
+  MoveContactWorker, 
+  MoveContactData, 
+  JobResult 
+} from '../../src/worker/move-contact-worker';
+
+describe('MoveContactWorker', () => {
+  let worker: MoveContactWorker;
+  let sandbox: SinonSandbox;
```
kennsippell, line 17: i've never used sinonsandbox but it is quite nice

src/worker/move-contact-worker.ts
```diff
@@ -0,0 +1,160 @@
...
+  private readonly MAX_CONCURRENCY = 1;         // Limit concurrency to 1 job at a time
+  private readonly MAX_SENTINEL_BACKLOG = 1000; // ensure we don't take down the server
```
kennsippell, line 22: roughly the number we have for `MAX_SENTINEL_BACKLOG` should be approximately how much of the sentinel backlog we expected to be processed within the `DELAY_IN_MILLIS`

[Here](https://github.com/moh-kenya/config-echis-2.0/issues/2174#issuecomment-2125502935) is a random single datapoint we can use. Looks like about 20,000 docs were processed in 3 hours.  So maybe let's do ~7000?

src/worker/move-contact-worker.ts
```diff
@@ -0,0 +1,160 @@
...
+      this.handleJob, 
+      { connection: redisConnection, concurrency: this.MAX_CONCURRENCY }
+    );
+  }
+
+  private handleJob = async (job: Job): Promise<boolean> => {
+    const jobData: MoveContactData = job.data;
+
+    // Ensure server availability
+    if (!(await this.canProcess(jobData))) {
```
kennsippell, line 40:
```suggestion
    if (await this.shouldPostpone(jobData)) {
```

? Slightly easier to read imo

src/worker/move-contact-worker.ts
```diff
@@ -0,0 +1,160 @@
...
+      await this.postpone(job);
+      return true;
+    }
+
+    const result = await this.moveContact(jobData);
+    if (!result.success) {
+      job.log(`[${new Date().toISOString()}]: ${result.message}`);
```
kennsippell, line 47: is there a way to just get timestamps on all log entries instead of doing it once here?

src/worker/move-contact-worker.ts
```diff
@@ -0,0 +1,160 @@
...
+      jobData: job.data,
+      jobOpts: {
+        ...job.opts,
+        delay: this.DELAY_IN_MILLIS
+      },
+      queueName: this.queueName,
+    };
+    await queueManager.addJob(jobParams);
+
+    const retryMessage = `Job postponed (not ready): ${job.id} (retry in ${this.DELAY_IN_MILLIS / 1000} sec.)`;
```
kennsippell, line 155:
```suggestion
    const retryMessage = `Job ${job.id} postponed until 9:35pm.  Reason was "sentinel backlog"`;
```

This would be my ideal. Do whatever you can or think is best

kennsippell (changes requested):
Great stuff Paul.  I did another review here but I got tired and didn't look through tests yet.

src/worker/move-contact-worker.ts
```diff
@@ -0,0 +1,160 @@
...
+      chtProcess.on('error', error => {
+        clearTimeout(timeout);
+        reject(error);
+      });
+    });
+  }
+
+  private async postpone(job: Job): Promise<boolean> {
+    // Add this job back to queue with a DELAY_IN_MILLIS delay
+    const jobParams = {
```
kennsippell, line 144: i tested this today and i'm unsure if it is working.  i had about 50 move-contact commands and they ran until sentinel backlog was >1000.  i waited well over an hour but they did not run again.

Dockerfile.worker
```diff
@@ -0,0 +1,15 @@
+FROM node:20-alpine
+
+WORKDIR /app
+
+ENV NODE_ENV production
+
+COPY package*.json .
+RUN apk add git
```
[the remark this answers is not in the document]
paulpascal: I just realized `cht-conf` actually requires `git` to be present in the cli env during installation

src/worker/move-contact-worker.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    console.log(`Job completed successfully: ${job.id}`);
+    return true;
+  };
+
+  private async canProcess(jobData: MoveContactData): Promise<boolean> {
+    try {
+      const { instanceUrl } = jobData;
+      const response = await axios.get(`${instanceUrl}/api/v2/monitoring`);
```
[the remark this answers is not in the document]
paulpascal: Interesting here @kennsippell 

I considered this approach, but I had concerns regarding the instanceUrl. The ChtApi class is designed to handle a single instanceUrl within its `Session/Axios instance`. However, in a deployment that supports multiple instances, the `Worker` could receive jobs from various instances. Since ChtApi only manages one instanceUrl at a time, I was afraid it wouldn’t be suitable for this use case where multiple instance URLs need to be processed concurrently.

Maybe I should investigate more.

src/worker/move-contact-worker.ts
paulpascal: I have a looked to that, but sorry could not find any better way to do this now.

paulpascal:
Hi, @mrjones-plip  it took a bit but finally got all the docker feedback addressed 🥲. Would you like to check it back when ever you have time?

@kennsippell, i have fixed the delay stuff as well.

Thanks

test/lib/move.spec.ts
```diff
@@ -27,11 +27,7 @@ describe('lib/move.ts', () => {
     const contactType = Config.getContactType('c_community_health_unit');
     const sessionCache = new SessionCache();
     
-    const actual = await MoveLib.move(formData, contactType, sessionCache, chtApi());
-    expect(actual.command).to.include('--contacts=chu-id');
```
kennsippell, line 31: should you make similar verifications on the data passed to mockQueueManager?

Dockerfile.worker
```diff
@@ -0,0 +1,15 @@
+FROM node:20-alpine
+
+WORKDIR /app
+
+ENV NODE_ENV production
```
mrjones-plip, line 5: I notice we hard code this to `production` in this file and in the main `Dockerfile`.  Can you not run `dev` mode in docker?  That seems a real loss!  Can we easily add that?

test/worker/move-contact-worker.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+import Auth from '../../src/lib/authentication';
+import { queueManager } from '../../src/shared/queues';
+import { 
+  MoveContactWorker, 
+  MoveContactData, 
+  JobResult 
+} from '../../src/worker/move-contact-worker';
+
+describe('MoveContactWorker', () => {
```
kennsippell, line 14: nit: generally the describe in this project has the path to the file rather than the name of a class. your call to follow

src/worker/move-contact-worker.ts
```diff
@@ -0,0 +1,156 @@
...
+      await this.postpone(job, processingToken);
+      throw new DelayedError();
+    }
+
+    const result = await this.moveContact(jobData);
+    if (!result.success) {
+      job.log(`[${new Date().toISOString()}]: ${result.message}`);
+      const errorMessage = `Job ${job.id} failed with the following error: ${result.message}`;
+      console.error(errorMessage);
+      throw new Error(errorMessage);
```
kennsippell, line 50: at some point it might be nice to handle errors  a bit better. maybe not in this PR, but what if cht-conf has an error - the most common scenario here is losing connectivity with the server during a heavy move. should the job be retried?


i'll make a quick comment about aborted/canceled jobs or generally handling any case where a contact's move is interrupted

* the good news is that if a move is interrupted, regardless of the state cht-conf can be rerun. it will aggregate all documents (those in the new position already and those in the old position) and move them all again. this is really an incredible win - if a job fails, it just needs to be retried.
* the bad news is that cht-user-management won't let you retry an aborted job if the top-level contact successfully moves. it will give an error like "A already has B as parent" or whatever. 

It's going to happen, and we're going to need to deal with it at some point.

kennsippell (changes requested):
Very very nice progress. This feels close.

env.example
```diff
@@ -1,7 +1,12 @@
...
-CHT_DEV_URL_PORT=localhost:5984 # where your dev CHT instance is, hostname:port
-CHT_DEV_HTTP=true              # 'false' for http  'true' for https
+COOKIE_PRIVATE_KEY= 
+QUEUE_PRIVATE_KEY=                                    # should be different from COOKIE_PRIVATE_KEY
+CONFIG_NAME=chis-ke                                   # Name of the configuration
+PORT=3000                                             # for development environmentcontainer)
+EXTERNAL_PORT=3000                                    # for docker
+INTERFACE=0.0.0.0                                     # Leave as '0.0.0.0' for prod, suggest '127.0.0.1' for development
+REDIS_HOST=localhost                                  # Redis server hostname use 'redis' for docker
+REDIS_PORT=6378                                       # Redis server port
```
mrjones-plip, line 8: suggest we leave these unset as both prod and dev will likely use the same  values.  As well, we should default to `redis` value (even though it's commented out):

```suggestion
#REDIS_HOST=redis                                  # Redis server hostname -  only uncomment if you know what you're doing
#REDIS_PORT=6378                                       # Redis server port -  only uncomment if you know what you're doing
```

mrjones-plip (changes requested):
yes!  Looking really good here. Most of my feedback is trying both make it easy to set up the dev environment, while also keeping sane defaults for production.

While I did easily stand up an instance using the bash script, I'm unable to connect to `redis` for some reason.

both the main service and the work service have this error, where `192.168.16.2` is the IP of the `redis` container:

```
docker logs cht-user-management-cht-user-management-1

Error: connect ECONNREFUSED 192.168.16.2:6378
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1606:16) {
  errno: -111,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '192.168.16.2',
  port: 6378
}
```

does this work for you?

src/liquid/place/move_form.html
```diff
@@ -46,3 +50,20 @@ <h1 class="subtitle">To</h1>
     </div>
   </div>
 </form>
+
+<script type="text/javascript">
+  function handleResponse(event) {
+    const response = event?.detail?.xhr?.responseText;
+    if (response && response.includes('data-success="true"')) {
```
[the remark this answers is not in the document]
paulpascal: Currently, it this isn't as expected, we will have an error (notification) displayed on top. But I added a console to log the failure.

paulpascal:
> yes! Looking really good here. Most of my feedback is trying both make it easy to set up the dev environment, while also keeping sane defaults for production.
> 
> While I did easily stand up an instance using the bash script, I'm unable to connect to `redis` for some reason.
> 
> both the main service and the work service have this error, where `192.168.16.2` is the IP of the `redis` container:
> 
> ```
> docker logs cht-user-management-cht-user-management-1
> 
> Error: connect ECONNREFUSED 192.168.16.2:6378
>     at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1606:16) {
>   errno: -111,
>   code: 'ECONNREFUSED',
>   syscall: 'connect',
>   address: '192.168.16.2',
>   port: 6378
> }
> ```
> 
> does this work for you?

Yes @mrjones-plip, it works for me, we can have a little sync on that if its okay with you

src/worker/move-contact-worker.ts
paulpascal: This is really important, the way I was dealing with it is to see from the `Board`, the issue with the failed job (from the logs).
If it has failed due to the lost of connectivity, then I just `retry` from the `Board`.

But we can handle this properly in another PR.

src/lib/authentication.ts
```diff
@@ -14,23 +16,42 @@ export default class Auth {
     if (!COOKIE_PRIVATE_KEY) {
       throw new Error('.env missing COOKIE_PRIVATE_KEY');
     }
+    if (!QUEUE_PRIVATE_KEY) {
+      throw new Error('.env missing QUEUE_PRIVATE_KEY');
```
kennsippell, line 20: can we also assert if COOKIE_PRIVATE_KEY and QUEUE_PRIVATE_KEY are equal?

test/worker/move-contact-worker.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    log: sinon.stub(),
+    moveToDelayed: sinon.stub(),
+  };
+
+  let shouldPostponeStub: sinon.SinonStub;
+  let moveContactStub: sinon.SinonStub;
+  let postponeStub: sinon.SinonStub;
+  let decodeTokenStub: sinon.SinonStub;
+
+  beforeEach(() => {
```
kennsippell, line 38: not required in this PR

i think a valuable test would be an integration test which calls `MoveLib.move()`, mocks the interactions with BullMQ (or which actually hosts redis during the test), and then verifies the `cht move-contacts` command which is kicked off by the runner

i love this sort of integration test and find it really value. the unit-style testing you have here (particularly tests focusing on private interfaces) are nice, but very sensitive to implementation change.  some integration tests or tests which are closer to user expectations would be great

kennsippell (changes requested):
Let's do one more, but I think we're basically there.

Dockerfile
```diff
@@ -2,18 +2,18 @@ FROM node:20-alpine
 
 ENV EXTERNAL_PORT 3000
 ENV PORT 3000
-ENV NODE_ENV production
 
 WORKDIR /app
 
 HEALTHCHECK --interval=30s --timeout=3s \
   CMD wget --spider http://localhost:${PORT}/_healthz || exit 1
```
mrjones-plip, line 9: My healthchecks were failing with this call b/c it was falling back to IPv6 (as seen in `docker inspect --format "{{json .State.Health }}" cht-user-management-cht-user-management-1 |jq`):

```
Connecting to localhost:3000 ([::1]:3000)
wget: can't connect to remote host: Connection refused\n
```

We can force it to use IPv4 and the healthcheck passes:

```suggestion
HEALTHCHECK --interval=30s --timeout=3s \
  CMD wget --spider http://127.0.0.1:${PORT}/_healthz || exit 1
```

mrjones-plip (changes requested):
Nice work @paulpascal! I REALLY love that `NODE_ENV` is exposed in the `.env` file now so it makes it trivial to run in dev mode in docker - so handy!

I'm not getting any more errors for `redis` service in the docker log, so I assume this is working? I don't know how to test.

I did have one issue that I wasn't able log in - here's what I did:
1. deploy `4.8.1` instance with docker helper `192-168-68-26.local-ip.medicmobile.org:10446`
2. set `CHT_DEV_HTTP=true` and `
CHT_DEV_URL_PORT=192-168-68-26.local-ip.medicmobile.org:10446` and `NODE_ENV=dev`
3. in CHT Core, create a `test` user that is role `User Manager`
4. start user man tool with: `./docker-local-setup.sh build`
5. try to login with that user on `127.0.0.1:3000/login`

expected: I can log in
actual: get error "login failed. please try again"

Here's my `.env` in case it's helpful:

```
NODE_ENV=dev
COOKIE_PRIVATE_KEY=quux
QUEUE_PRIVATE_KEY=smang
CONFIG_NAME=chis-ke
PORT=3000
EXTERNAL_PORT=3000
INTERFACE=0.0.0.0
CHT_DEV_HTTP=true
CHT_DEV_URL_PORT=192-168-68-26.local-ip.medicmobile.org:10446
```

README.md
```diff
@@ -149,3 +154,18 @@ Docker images are hosted on [AWS's Elastic Container Registry (ECR)](https://gal
...
+
+```bash
+./docker-local-setup.sh build
+```
+
+If you just need to run the development environment without rebuilding the images run:
+
+```bash
+./docker-local-setup.sh
+```
```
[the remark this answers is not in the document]
mrjones-plip: moved content looks great - thank you!

mrjones-plip (approved):
From a docker and docker compose perspective, with the updated `env.example` and updated `readme.md` - I think this PR is ready to go from my perspective.  I still was unable to log in, so please be sure and test this with a clean set up to ensure there's no regressions!

paulpascal:
> ... I still was unable to log in, so please be sure and test this with a clean set up to ensure there's no regressions!

Thanks @mrjones-plip , I will make sure to have that working.

src/server.ts
```diff
@@ -44,6 +45,7 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
   });
 
   Auth.assertEnvironmentSetup();
+  checkRedisConnection();
```
kennsippell, line 48: Feels like this should await somehow? Is the idea that it will crash the app whenever it completes async? Does it work?

test/scripts/wait_for_redis.sh
kennsippell, line 1: Sorry, I don't see where this is used?

kennsippell (approved):
Amazing! Let's go!

test/scripts/wait_for_redis.sh
paulpascal: Oh yeah , It's used in the script to run the integration test,

package.json/scripts

src/server.ts
paulpascal: Yes exactly, the idea is for it to crash whenever it completes and it works.

PR #178, opened 2024-05-22
Script to bulk move contacts from CSV

kennsippell:
Based on #177. Work done for #12. 

https://github.com/moh-kenya/config-echis-2.0/issues/2174 required me to move 381 places within the Laikipia County. I wrote a quick script and executed it using @paulpascal 's worker thread. Test notes captured [here](https://github.com/medic/cht-user-management/pull/177#issuecomment-2123592729).


scripts/move-contacts/index.ts
```diff
@@ -0,0 +1,34 @@
...
+import ChtSession from '../../src/lib/cht-session';
+import MoveLib from '../../src/lib/move';
+import SessionCache from '../../src/services/session-cache';
+
+import allPlacesToMove from './laikipia-west.json';
+
+const username = 'kenn_sippell_medic_user_manager';
+const password = '';
+const authInfo = Config.getAuthenticationInfo('laikipia.echis.go.ke');
+const contactType = Config.getContactType('d_community_health_volunteer_area');
```
derickl, line 13: This is a very neat solution especially given the re-use of components from he user management tool. We should definitely look at having an experience built out for this and make it accessible to a few user managers and learn with them.

Nice work @kennsippell

derickl: and @paulpascal for the foundational work

kennsippell: I think cht-user-management is made in a way such that it is a powerful scripting library for custom user management work.

scripts/move-contacts/laikipia-west.json
```diff
@@ -0,0 +1,457 @@
+[
```
derickl, line 1: you probably need to include a csv / xlsx to json converter

kennsippell: ya this was just quick and dirty. i wasn't planning to check this in less people say [here](https://medic.slack.com/archives/C066J5C8Z88/p1716338933731549) that they think it's useful. so far nobody said anything, so I'll just discard this

PR #183, opened 2024-06-20
feat(#169): cache expiry for remote places

### LINKED ISSUE #169: Cache expiry for remote places
This tool downloads and caches the places on an instance to make search functionality fast. The cache currently never expires and relies only on manual refresh. There should likely be some sort of automated expiry so users need to worry about this less.

inromualdo:
closes #169


kennsippell:
I'm so pleased you're working on this.  There is quite a bit of confusion from people in Kenya when the cache is stale.  Much appreciated @inromualdo

inromualdo:
Hi @kennsippell 

Here’s the approach I’m taking to resolve the issue:

- I’ve integrated the node-cache package to handle caching.

- The primary change is in the cht-api file, where I initialize the cache with a one-hour expiration time.

- When retrieving data for a place by type, if the data is not in the cache, an Axios request is made; otherwise, the cached content is returned.

- Whenever a place of a given type is updated or created, the cache for that type is cleared to ensure that the next read request fetches fresh data from the API.

- Consequently, the remote-place-cache file is no longer necessary, which has resulted in changes across various files and unit tests.

Currently, some tests are failing. I am working on resolving these issues and performing a manual test. In the meantime, please review my progress and provide any feedback or suggestions you may have on the process.

src/liquid/app/nav.html
```diff
@@ -47,9 +47,6 @@
       
         <div class="navbar-dropdown">
           {% if op == 'table' %}
-            <a class="navbar-item" hx-post="/app/refresh-all" hx-swap="none">
```
kennsippell, line 50: Should we not support manual refresh? Do you need to wait an hour?

src/lib/cht-api.ts
```diff
@@ -182,20 +189,28 @@ export class ChtApi {
       include_docs: true,
     };
     console.log('axios.get', url, params);
-    const resp = await this.axiosInstance.get(url, { params });
+    let places: any[] | undefined = ChtApi.cache.get(placeType);
```
kennsippell, line 192: previously cache was keyed by `[domain][contact-type]` but now only by `[contact-type]`.  Isn't the domain dimension required? This allowed independant caches and timers for each county eCHIS instance... just because the cache has expired for Nairobi county instance, should it also expire for Migori county instance?

src/lib/cht-api.ts
```diff
@@ -182,20 +189,28 @@ export class ChtApi {
...
+    if (places === undefined) {
+      const resp = await this.axiosInstance.get(url, { params });
+      places = resp.data.rows;
+      ChtApi.cache.set(placeType, places);
+    }
 
-    return resp.data.rows
-      .map((row: any): RemotePlace => {
+    return places?.map((row: any): RemotePlace => {
```
kennsippell, line 199: welcome you to implement the cache via the `remote-place-cache` class but using this node-cache library.  everything in ChtApi isn't easily unit testable, but having this cache in a separate class allows it to be continue to be tested

kennsippell (changes requested):
nice ideas here. some feedback

inromualdo:
@kennsippell , I'm replacing this PR by [PR#237](https://github.com/medic/cht-user-management/pull/237).

Issue #203, opened 2024-09-13
CIV config updates


aloundoye:
- [x] add hints on CIV configs
- [x] remove email and username on the config

PR #204, opened 2024-09-13
fix(#203): update civ config

aloundoye:
#203 
- add hints on CIV configs
- remove email and username on the config(don't appear on the contact form)


inromualdo (approved):
LGTM

PR #207, opened 2024-09-23
add support for multiple facility ids

freddieptf:
User details in cht 4.10.0 have `facility_id` as an array of ids. This PR updates the implementation to support that so scoping can still work correctly

### DIFF
--- a/src/lib/cht-session.ts
+++ b/src/lib/cht-session.ts
@@ -52,7 +52,12 @@ export default class ChtSession {
   }
 
   public static createFromDataString(data: string): ChtSession {
-    const parsed: { authInfo: AuthenticationInfo, sessionToken: string, username: string, facilityIds: string[] } = JSON.parse(data);
+    const parsed: { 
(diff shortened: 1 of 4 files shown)


paulpascal (approved):
Awesome @freddieptf, LGTM

src/lib/cht-session.ts
```diff
@@ -99,10 +104,13 @@ export default class ChtSession {
...
-    return {
-      isAdmin,
-      facilityId: resp.data?.facility_id,
-    };
+    let facilityId;
+    if (typeof resp.data?.facility_id === 'string') {
+      facilityId = [resp.data.facility_id];
+    } else if (Array.isArray(resp.data?.facility_id)) {
+      facilityId = resp.data.facility_id;
+    }
```
kennsippell, line 112: else? throw?

freddieptf: I didn't add it as there's an already existing null-check and throw where this function is called

Issue #198, opened 2024-08-28
Integrate the user management tool with Mali-CHW


aloundoye:
We are planning to integrate the user management tool with Mali-CHW

PR #208, opened 2024-09-30
Add MoH Mali-CHW config

paulpascal:
#198


paulpascal:
Handled in PR [292](https://github.com/medic/cht-user-management/pull/264)

PR #209, opened 2024-10-04
add support for multiple facility ids (#207)

freddieptf:
* users can now be associated to multiple facilities

* 1.4.1

### DIFF
--- a/src/lib/cht-session.ts
+++ b/src/lib/cht-session.ts
@@ -15,14 +15,14 @@ axiosRetry(axios, axiosRetryConfig);
 export default class ChtSession {
   public readonly authInfo: AuthenticationInfo;
   public readonly username: string;
-  public readonly facilityId: string;
+  public readonly facilityIds: string[];
   public readonly axiosInstance: AxiosInstance;
   public readonly sessionToken: string;
 
-  private constructor(authInfo: AuthenticationInfo, sessionToken: string, username: string, facilityId: string) {
+  private constructor(authInfo: AuthenticationInfo, sessionToken: string, username: string, facilityIds: string[]) {
     this.authInfo = authInfo;
     this.username = username;
-    this.facilityId = facilityId;
+    this.facilityIds = facilityIds;
     this.sessionToken = sessionToken;
     
     this.axiosInstance = axios.create({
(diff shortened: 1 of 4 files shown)


src/lib/cht-session.ts
```diff
@@ -15,23 +15,22 @@ axiosRetry(axios, axiosRetryConfig);
 export default class ChtSession {
   public readonly authInfo: AuthenticationInfo;
   public readonly username: string;
-  public readonly facilityId: string;
+  public readonly facilityIds: string[];
   public readonly axiosInstance: AxiosInstance;
   public readonly sessionToken: string;
 
-  private constructor(authInfo: AuthenticationInfo, sessionToken: string, username: string, facilityId: string) {
+  private constructor(authInfo: AuthenticationInfo, sessionToken: string, username: string, facilityIds: string[]) {
```
kennsippell, line 22: won't this be a `string` and not a `string[]` on older versions of cht-core?

src/lib/cht-session.ts
```diff
@@ -44,25 +43,30 @@ export default class ChtSession {
...
   public static createFromDataString(data: string): ChtSession {
-    const parsed:any = JSON.parse(data);
-    return new ChtSession(parsed.authInfo, parsed.sessionToken, parsed.username, parsed.facilityId);
+    const parsed: { 
+      authInfo: AuthenticationInfo;
+      sessionToken: string;
+      username: string;
+      facilityIds: string[]; 
+    } = JSON.parse(data);
+    return new ChtSession(parsed.authInfo, parsed.sessionToken, parsed.username, parsed.facilityIds);
```
kennsippell, line 61: isn't it `facilityId` on old versions of core?  

seems we need two tests for this PR. One with each of the values returned by cht-core

src/lib/cht-session.ts
kennsippell: my point about the test stands please. it's hard to keep these scenarios straight without regression tests

src/lib/cht-session.ts
kennsippell: don't we need to invalidate all existing sessions? since you're changing the values which are stored? anybody who is logged in will fail this code path

kennsippell:
Why is base branch for this `deploy`? Sorry if I'm behind on what's happening here.

freddieptf:
Base branch had to change because of https://github.com/medic/helm-charts/issues/19

src/lib/cht-session.ts
freddieptf: I think this Is something that has to be done via kubectl https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/#edit-secret. Not sure if there's a way to regenerate it without deleting then creating it again

PR #210, opened 2024-10-04
Without move contact feature

paulpascal:


paulpascal:
No more need since the helm charts templates have been nice handled.

Latest changes can be easily deployed.

cc: @freddieptf

Issue #192, opened 2024-07-24
User unable to move CHA to another CHU


IreneAchola:
There is some confusion when moving CHA(CHU?)

- It is unclear whether the process moves a CHA or a CHU, since the heading changes on selecting Community Health Unit as shown in image 2.
- In the case where the user is moving a CHA, it does not allow the user to specify the destination CHU, only the subcounty.
- The name of CHA being moved is not indicated.
- The breadcrumb is misleading. As on image 3, the before indicates `old CHU >> old subcounty`, while the after indicates `old subcounty >> new subcounty`
<img width="680" alt="Screenshot 2024-07-24 at 11 16 43" src="https://github.com/user-attachments/assets/ba4435c0-d390-4301-acad-cc128cd1d9ec">
<img width="680" alt="Screenshot 2024-07-24 at 11 25 41" src="https://github.com/user-attachments/assets/e4505018-15a8-40a9-a68a-d5472a2e9567">
<img width="680" alt="Screenshot 2024-07-24 at 11 39 53" src="https://github.com/user-attachments/assets/60c4f9d5-c02c-44cb-ab76-fb48ddcad076">

alexosugo:
@PhilipNgari Updates have largely been pushed by individual devs targeting specific needs/deployments over the last few months. There is no specific lead on the UMT.

billwambua:
@IreneAchola in your example it moves the CHU with the CHA contact as well of course. 

But it does not move the CHA only as the title suggests.

There is no way to move only the CHA with this tool it seems.

Issue #139, opened 2024-04-17
Rename "Replace CHU" to "Replace CHA"


kennsippell:
> The label should be Replace Existing CHA. The confusion came in as they thought this would create a new CHU and then move CHP areas to the new CHU.
> Ideally we should only have the subcounty, chu and CHA Name/Phone

kennsippell:
![image](https://github.com/medic/cht-user-management/assets/9014751/9e74ad4e-7072-4dbe-bbcf-1814f39bd82e)

Reactivating this since we are now getting different feedback.

PhilipNgari:
weighing on this too as user managers have the same challenge.let us rename it to  "community health Assistant" as that's the expected action

additionally, the new CHP details section below outgoing CHP should have a label e.g Incoming CHP

PR #211, opened 2024-10-23
Fix form labels

freddieptf:
Fixes https://github.com/medic/cht-user-management/issues/139
Fixes https://github.com/medic/cht-user-management/issues/192


paulpascal (approved):
Nicely done @freddieptf 
LGTM ✨

PR #212, opened 2024-10-29
error if user tries to create a place with the defined dedup property in use

### LINKED ISSUE #181: Users should not be able to create duplicate places
There's a case in Makueni where a user manager created a duplicate CHU in the same subcounty. I would not expect this to be possible.

Affected CHU - Kilala Community Health Unit
One was created on Feb 5th 2024 and the other was created on May 18th 2024

freddieptf:
Fixes #181 

This adds a way to prevent a user from creating duplicate places by defining a `dedup_property` on a place. The `dedup_property` is not globally unique but it is scoped to the parent place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Property | Type | Description
 `contact_types.contact_properties` | Array<ConfigProperty> | Defines the attributes which are collected and set on the user's primary contact doc. See [ConfigProperty](#ConfigProperty).
 `contact_types.deactivate_users_on_replace` | boolean | Controls what should happen to the defunct contact and user documents when a user is replaced. When `false`, the contact and user account will be deleted. When `true`, the contact will be unaltered and the user account will be assigned the role `deactivated`. This allows for account restoration.
 `contact_types.hint` | string | Provide a brief hint or description to clarify the expected input for the property.
+`contact_types.dedup_property` | string | The place's `property_name` used to prevent creation of duplicates under a certain parent.
 `logoBase64` | Image in base64 | Logo image for your project
 
 #### ConfigProperty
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -48,6 +48,7 @@ Property | Type | Description
 `contact_types.contact_properties` | Array<ConfigProperty> | Defines the attributes which are collected and set on the user's primary contact doc. See [ConfigProperty](#ConfigProperty).
 `contact_types.deactivate_users_on_replace` | boolean | Controls what should happen to the defunct contact and user documents when a user is replaced. When `false`, the contact and user account will be deleted. When `true`, the contact will be unaltered and the user account will be assigned the role `deactivated`. This allows for account restoration.
 `contact_types.hint` | string | Provide a brief hint or description to clarify the expected input for the property.
+`contact_types.dedup_property` | string | The place's `property_name` used to prevent creation of duplicates under a certain parent.
```
kennsippell, line 51: [This PR](https://github.com/medic/cht-user-management/pull/167) which introduced a warning system also proposed a [config schema](https://github.com/medic/cht-user-management/pull/167/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R61) which enforced uniqueness (or "dedupes"). 

I still would love to get that merged some day. It was derailed by https://github.com/medic/cht-user-management/commit/becba2552d8dbb5ad1730cffacf62ebd9534edc7 and I never did the work to update it.  

This seems like a partial implementation of features within that PR. It sure would be great to use something like the same config schema so that PR can eventually make it in.

freddieptf: oh i forgot about that, i'll have a look at both https://github.com/medic/cht-user-management/pull/161 and https://github.com/medic/cht-user-management/pull/167 today

freddieptf:
Fixed in #167

PR #213, opened 2024-10-30
updated values for users-tg

Hareet:
New values for users-chis-tg on moh-togo.app.

Can be used as an example values file.
Please note REDIS_HOST naming convention, and URL variables

### DIFF
--- a/scripts/deploy/values/users-chis-tg.yaml
+++ b/scripts/deploy/values/users-chis-tg.yaml
@@ -1,8 +1,29 @@
 cht-user-management:
+  replicaCount: 1
+  service:
+    port: 3000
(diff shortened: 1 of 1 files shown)


mrjones-plip (changes requested):
I'm still very rusty on helm works, so not really checking the structure/logic to any of this :/

That said, we [have 4 values files](https://github.com/medic/cht-user-management/tree/main/scripts/deploy/values) in this repo and the're all 1:1 used for all 4 deployments.  Makes sense to update the other 3 while we're in here.

Hareet:
Closing.

@mrjones-plip Paul got all the configs covered in another PR

PR #219, opened 2024-11-01
change .env to use http by default

mrjones-plip:
oops! `CHT_DEV_HTTP` should match the value in `CHT_DEV_URL_PORT`, so it should be `true` (don't use TLS).  I changed his by accident in my last PR!


paulpascal (approved):
Oops,
Albright

Issue #193, opened 2024-07-25
eCHIS Kenya: Support for moving households


derickl:
There are cases of CHPs manning several villages due to a shortage of CHPs. When a new one is appointed to take over a village, part of these registered households need to be reallocated to the new CHP.

kennsippell:
Today there was also a request from Lamu to move all households for a CHP area into another. It is almost merging the contacts, but the source CHP area should remain.

derickl:
@freddieptf what is blocking this issue?

PR #227, opened 2024-11-19
Optionally enable/disable features on contact types

freddieptf:
This PR introduces a way to explicitly enable/disable the main 3 features of this app i.e creating places, replacing contacts and moving places. This came from [this request](https://github.com/medic/cht-user-management/issues/193) where users need to be allowed to move households **but not** create/replace. All features are enabled by default but this can be configured using the `feature_flags` field.

### DIFF
--- a/src/config/chis-ke/config.json
+++ b/src/config/chis-ke/config.json
@@ -188,7 +188,6 @@
       "friendly": "West Pokot",
       "domain": "westpokot.echis.go.ke"
     },
-
     {
       "friendly": "Staging (chis-staging.health.go.ke)",
       "domain": "chis-staging.health.go.ke"
@@ -335,6 +334,41 @@
           "parameter": "KE",
           "type": "phone",
           "required": true
+        }
+      ]
+    },
+    {
+      "name": "e_household",
+      "friendly": "Household",
+      "contact_type": "f_client",
+      "user_role": [],
+      "place_properties": [],
+      "contact_properties": [],
+      "feature_flags": [
+        "move"
+      ],
+      "replacement_property": {
+        "friendly_name": "",
+        "property_name": "replacement",
+        "type": "name",
+        "required": true
+      },
+      "hierarchy": [
+        {
+          "friendly_name": "CHU",
+          "property_name": "CHU",
+          "contact_type": "c_community_health_unit",
+          "type": "name",
+          "required": true,
+          "level": 2
+        },
+        {
+          "friendly_name": "CHP Area",
+          "property_name": "CHP",
+          "contact_type": "d_community_health_volunteer_area",
+          "type": "name",
+          "required": true,
+          "level": 1
         }
       ]
     }
--- a/src/config/chis-ke/config.json
+++ b/src/config/chis-ke/config.json
@@ -344,9 +344,10 @@
       "user_role": [],
       "place_properties": [],
       "contact_properties": [],
-      "feature_flags": [
-        "move"
-      ],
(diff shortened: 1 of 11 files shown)


paulpascal (changes requested):
Very nice @freddieptf 🍞

I left some little feedback inline

src/config/index.ts
```diff
@@ -19,13 +19,14 @@ export type ContactType = {
...
-  username_from_place: boolean;
+  username_from_place?: boolean;
   hierarchy: HierarchyConstraint[];
   replacement_property: ContactProperty;
   place_properties: ContactProperty[];
   contact_properties: ContactProperty[];
-  deactivate_users_on_replace: boolean;
+  deactivate_users_on_replace?: boolean;
   hint?: string;
+  feature_flags?: string[];
```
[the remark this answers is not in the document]
freddieptf: I've added an enum with the expected values, the config should fail parsing if it's not part of the enum values

src/config/chis-ke/config.json
```diff
@@ -337,6 +336,41 @@
           "required": true
         }
       ]
+    },
+    {
+      "name": "e_household",
+      "friendly": "Household",
+      "contact_type": "f_client",
+      "user_role": [],
```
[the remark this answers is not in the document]
freddieptf: Fixed

src/config/config-factory.ts
```diff
@@ -4,11 +4,43 @@ import kenyaConfig from './chis-ke';
...
+  return {
+    config: {
+      ...c.config,
+      contact_types: c.config.contact_types.map((t: any) => {
+        return {
+          ...t,
+          feature_flags: t.feature_flags?.map((v: string) => {
+            if ((Object.values(Feature) as string[]).indexOf(v) === -1) {
+              throw new Error(
+                'invalid feature flag: ' +
```
kennsippell, line 23: Can we use a config syntax which doesn't need parsing? What about three booleans: `can_edit`, `can_replace` and `can_move`?

src/config/index.ts
```diff
@@ -19,13 +19,14 @@ export type ContactType = {
   contact_type: string;
   contact_friendly?: string;
   user_role: string[];
-  username_from_place: boolean;
+  username_from_place?: boolean;
```
kennsippell, line 22: why not require these to have explicit values? seems cleaner to be explicit instead of having to worry about defaults

src/lib/remote-place-resolver.ts
```diff
@@ -111,6 +112,7 @@ export default class RemotePlaceResolver {
 
 function getFuzzFunction(place: Place, hierarchyLevel: HierarchyConstraint, contactType: ContactType) {
   const fuzzingProperty = hierarchyLevel.level === 0 ? contactType.replacement_property : hierarchyLevel;
+  assert(fuzzingProperty);
```
kennsippell, line 115: Is this change intended to be committed?

kennsippell (changes requested):

kennsippell:
Have you tried this change on production data like the Nairobi instance. It will download and cache every household... Does it just explode?

src/config/config-factory.ts
freddieptf: Yeah that's more readable, wanted to have a shorter config file but might not be worth it

freddieptf:
I tried it on production and the request never completed so we can't actually search for households this way. Not sure if using the `contacts_by_parent` view help resolve the places faster

PR #237, opened 2025-01-06
feat(#169): implement cache expiry for remote places

### LINKED ISSUE #169: Cache expiry for remote places

inromualdo:
Replace PR#183

closes #169 

Note: Failing unit test will be fix by [PR#239](https://github.com/medic/cht-user-management/pull/239)


src/lib/remote-place-cache.ts
```diff
@@ -24,52 +18,56 @@ export type RemotePlace = {
...
+    if (!this.cache) {
+      this.cache = new NodeCache({ 
+        stdTTL: 3600,
+        checkperiod: 120 
+      });
+    }
+    return this.cache;
+  }
+
+  private static getCacheKey(domain: string, placeType?: string): string {
```
kennsippell, line 47: under what situation should the `placeType` be undefined? i think we should probably assert in that case, rather than supporting it

src/lib/remote-place-cache.ts
```diff
@@ -24,52 +18,56 @@ export type RemotePlace = {
...
-    const places = domainCache[domain]?.[placeType];
-    if (!places) {
-      const fetchPlacesWithType = RemotePlaceCache.fetchRemotePlaces(chtApi, contactType, hierarchyLevel);
-      domainCache[domain] = {
-        ...domainCache[domain],
-        [placeType]: await fetchPlacesWithType,
-      };
+      this.getCache().flushAll();
+    } else {
+      const cacheKey = this.getCacheKey(domain, contactTypeName);
```
kennsippell, line 68: can you test the scenario where you `clear(domain)` without any placeType? This should clear all of the place types linked to that domain -- but I think what is happening here is that none of them are cleared.

maybe add a test for that?

kennsippell (changes requested):
Nice! I think this is really valuable work.

src/lib/remote-place-cache.ts
```diff
@@ -31,8 +24,38 @@ export type RemotePlace = {
 };
 
 export default class RemotePlaceCache {
-  private static cache: RemotePlaceDatastore = {};
+  private static cache: NodeCache;
+  private static readonly CACHE_TTL = 3600;
```
kennsippell, line 28: is this one hour? i suspect we want it to be like 24 hrs? can you share your thinking? for nairobi i think this takes over a minute and 1 hour expiration would guarantee basically users do it once per session. is that what we want?

src/lib/remote-place-cache.ts
```diff
@@ -31,8 +24,38 @@ export type RemotePlace = {
...
+    let places = this.getCache().get<RemotePlace[]>(cacheKey);
+    if (!places) {
+      places = await this.getRemotePlaces(chtApi, contactType, hierarchyLevel);
+      this.getCache().set(cacheKey, places);
+    }
+    return places;
+  }
+
+  private static getCache(): NodeCache {
+    if (!this.cache) {
```
kennsippell, line 46: it seems easy to use RemotePlaceCache.cache instead of getCache? Is to  make that harmless? Like calling `initialize` once `RemotePlaceCache.initialize()` once within the module scope and then you can use `this.cache` safely everywhere else

src/lib/remote-place-cache.ts
```diff
@@ -31,8 +24,38 @@ export type RemotePlace = {
 };
 
 export default class RemotePlaceCache {
-  private static cache: RemotePlaceDatastore = {};
+  private static cache: NodeCache;
+  private static readonly CACHE_TTL = 3600;
+  private static readonly CACHE_CHECK_PERIOD = 120;
+  
+  public static async getPlacesWithType(chtApi: ChtApi, contactType: ContactType, hierarchyLevel: HierarchyConstraint)
```
kennsippell, line 31: what the difference between `public getPlacesWithType` and `public getRemotePlaces` in this class? are two interfaces like this required?

src/lib/remote-place-cache.ts
```diff
@@ -47,23 +70,35 @@ export default class RemotePlaceCache {
...
-      delete this.cache[domain][contactTypeName];
+      this.getCache().flushAll();
+      return;
+    }
+
+    if (!contactTypeName) {
+      // Clear all keys matching domain prefix
+      const keys = this.getCache().keys();
+      const domainPrefix = `${domain}:`;
+      keys.forEach(key => {
```
kennsippell, line 94:
```suggestion
      keys.filter(key => key.startsWith(domainPrefix))
        .forEach(this.cache.del)
```

Consider ?

src/lib/remote-place-cache.ts
```diff
@@ -47,23 +70,35 @@ export default class RemotePlaceCache {
   public static add(place: Place, chtApi: ChtApi): void {
     const { domain } = chtApi.chtSession.authInfo;
     const placeType = place.type.name;
-
-    const places = this.cache[domain]?.[placeType];
-    // if there is no cache existing, discard the value
-    // it will be fetched if needed when the cache is built
-    if (places) {
+    const cacheKey = this.getCacheKey(domain, placeType);
```
kennsippell, line 73: lets say there is a place `x` of type `type` on the instance. 

scenario 1
1. the cache is empty
2. run get('type') 
you'll get x

but scenario 2
1. the cache is empty
2. add(y) of `type`
3. get('type') 

now you won't see x.  i think this is a bug, no?

src/lib/remote-place-cache.ts
```diff
@@ -47,23 +70,35 @@ export default class RemotePlaceCache {
...
+    
+    const places = this.getCache().get<RemotePlace[]>(cacheKey) || [];
+    
+    if (places.length === 0) {
```
kennsippell, line 77: why do you only add if places is empty?  previously, we only added if not empty...

src/lib/remote-place-cache.ts
```diff
@@ -118,4 +152,11 @@ export default class RemotePlaceCache {
 
     return [];
   }
+
+  // For testing purposes
+  public static hasData(domain: string, placeType: string): boolean {
```
kennsippell, line 157: you can just access private members in tests. shouldn't this be a function within the spec.ts file?

kennsippell (changes requested):
nice!

src/lib/remote-place-cache.ts
inromualdo: I initially set it to 1h as a default to balance data freshness with performance. However, with nairobi scenario where fetching places takes over a minute, a longer TTL would make more sense. I agree that extending it to 24h would be more appropriate for this use case. Alternatively, we could make it adaptive by adding it to the configuration options, allowing developers to set custom values (8h, 12h, etc.) based on deployment needs...

src/lib/remote-place-cache.ts
inromualdo: This approach makes the code harmless, as it guarantees the cache is always initialized before use, centralizes the initialization logic, and prevents having to repeat null checks throughout the code.

src/lib/remote-place-cache.ts
```diff
@@ -31,8 +24,24 @@ export type RemotePlace = {
 };
 
 export default class RemotePlaceCache {
-  private static cache: RemotePlaceDatastore = {};
+  private static cache: NodeCache;
```
kennsippell, line 27:
```suggestion
  private static cache: NodeCache = new NodeCache({
      stdTTL: this.CACHE_TTL,
      checkperiod: this.CACHE_CHECK_PERIOD 
  });
```

Can get rid of getCache() if you do this inline

src/lib/remote-place-cache.ts
```diff
@@ -47,23 +56,35 @@ export default class RemotePlaceCache {
...
-      delete this.cache[domain][contactTypeName];
+      this.getCache().flushAll();
+      return;
+    }
+
+    if (!contactTypeName) {
+      // Clear all keys matching domain prefix
+      const keys = this.getCache().keys();
+      const domainPrefix = `${domain}:`;
+      keys.filter(key => key.startsWith(domainPrefix)).forEach(key => this.getCache().del(key));
```
kennsippell, line 84: break to multiple lines?

src/lib/remote-place-cache.ts
```diff
@@ -72,17 +93,16 @@ export default class RemotePlaceCache {
...
     const { domain } = chtApi.chtSession.authInfo;
     const placeType = hierarchyLevel.contact_type;
-    const { cache: domainCache } = RemotePlaceCache;
-    const places = domainCache[domain]?.[placeType];
-    if (!places) {
+    const cacheKey = this.getCacheKey(domain, placeType);
+    
+    const cacheData = this.getCache().get<RemotePlace[]>(cacheKey);
+    if (!cacheData) {
       const fetchPlacesWithType = this.fetchRemotePlacesAtLevel(chtApi, hierarchyLevel);
```
kennsippell, line 100: The reason this was split before is to cover the scenario:  

1. Invoke `fetchCachedOrRemotePlaces`
2. Before the above call completes, call `fetchRemotePlacesAtLevel` again.

Expect: Both calls should not invoke `fetchRemotePlacesAtLevel` because it is very heavy and slow. The second call should be blocked on the result of the first.

I think that behavior has regressed. Can you add a test for it and fix?

kennsippell (changes requested):

src/lib/remote-place-cache.ts
```diff
@@ -72,17 +90,34 @@ export default class RemotePlaceCache {
...
+    }
+
+    // Check if a fetch is already in progress
+    if (this.runningFetch.has(cacheKey)) {
+      return this.runningFetch.get(cacheKey)!;
     }
 
-    return domainCache[domain][placeType];
+    // Initiate fetch and store the promise
+    const fetchPromise = (async () => {
```
kennsippell, line 107: can we split this

```
async function fetchPlaces() {
};
const promiseToFetch = fetchPlaces();
```

kennsippell (approved):
nice!

PR #240, opened 2025-01-14
Makes configuration uploadable thru API

### LINKED ISSUE #194: separate config from main source code
Like CHT config needs to be separate from main code (`src/config/config-factory.ts`) this will make configuration more flexible

jonathanbataire:
## Description
Current app requires rebuilding and re-deploying a app in order to change config
This approach introduces uploadable config making it easier to change configuration 
The app defaults to the prebuilt config, once a new configuration is uploaded it overrides any pre-built config
Configuration can only be updated by CHT admins
Fixes #194 

### changes
- introduces new api end point `/api/config` used with basic auth to upload config

```
curl -X POST https://username:password@user-management.com/api/config?domain=cht-instance.com \
     -H "Content-Type: application/json" \
     -d @config.json
```


src/lib/cht-session.ts
freddieptf, line 1: Is this change necessary?

jonathanbataire: yes, isAdmin is more of a chtSession property than an action on chtSession

src/server.ts
```diff
@@ -63,6 +63,32 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
       return;
     }
 
+    const authHeader = req.headers.authorization as string;
```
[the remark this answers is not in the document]
jonathanbataire: Basic authentication for the API
currently we only have cookie/session based auth

src/server.ts
```diff
@@ -71,6 +97,10 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
 
     try {
       const chtSession = Auth.createCookieSession(cookieToken);
+      if (req.routeOptions.url === '/api/config' && !chtSession?.isAdmin) {
```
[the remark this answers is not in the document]
jonathanbataire: I think authorization should be handled in the auth hook/middleware 
but i have no strong opinion about this though i can move it

src/config/config-factory.ts
```diff
@@ -1,24 +1,64 @@
...
+    const jsonString: string = JSON.stringify(jsonConfigData, null, 2);
+    await fs.promises.writeFile(uploadedConfigFilePath, jsonString);
+  } catch (error) {
+    throw new Error('writeConfig: Failed to write file');
+  }
+}
+
+export async function readConfig(): Promise<PartnerConfig> {
+  try {
+    const fileContent = await fs.promises.readFile(uploadedConfigFilePath, 'utf-8');
```
freddieptf, line 56: I see alot of calls had to be converted to async so i have to ask, Is there a synchronous equivalent to this?

jonathanbataire: yeah sure there is but its asynchronous mainly for performance.
since this is an api/server you'd want to support concurrency with multiple requests without blocking
also config is read almost every time you make a request to the server for a page

freddieptf: If It's for performance, wouldn't caching the config be more performant than reading from disk for every request?

jonathanbataire: maybe maybe not
but generally i wouldn't advice reading file synchronously in a server cos of its blocking nature
It more like reading db data synchronously, not advisable

src/server.ts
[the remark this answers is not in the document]
jonathanbataire: i have added a config factory class
Ready for review @freddieptf

src/server.ts
```diff
@@ -63,6 +63,32 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
       return;
     }
 
+    const authHeader = req.headers.authorization as string;
+    if (authHeader && authHeader.startsWith('Basic ')) {
+      if (!req.routeOptions.url?.startsWith('/api')) {
```
freddieptf, line 68: When i POST the config without credentials it will try to redirect me to the `/login` endpoint which i think is awkward for an API. Is there any way this can be improved?

src/config/index.ts
```diff
@@ -139,19 +138,27 @@ export class Config {
...
   }
 
-  public static getAuthenticationInfo(domain: string) : AuthenticationInfo {
-    const domainMatch = Config.getDomains().find(c => c.domain === domain);
+  public static getAuthenticationInfo(domain: string): AuthenticationInfo {
+    const availableDomains = Config.getDomains();
+
+    //if config has default config i.e a new install
+    //return passed domain for api config upload authentication
+    if (availableDomains.some(c => c.domain === 'localhost')) {
```
freddieptf, line 146: I don't understand the comment here, why are you explicitly checking for localhost?

src/config/index.ts
```diff
@@ -139,19 +138,27 @@ export class Config {
...
+
+    const domainMatch = availableDomains.find(c => c.domain === domain);
     if (!domainMatch) {
       throw new Error(`unrecognized domain: "${domain}"`);
     }
     return domainMatch;
   }
 
-  public static getLogoBase64() : string {
+  public static async getLogoBase64(): Promise<string> {
```
freddieptf, line 157: Doesn't need to be async

src/routes/add-place.ts
```diff
@@ -34,7 +34,8 @@ export default async function addPlace(fastify: FastifyInstance) {
 
   fastify.post('/place/dob', async (req, resp) => {
     const { place_type, prefix, prop_type } = req.query as any;
-    const contactType = Config.getContactType(place_type).contact_properties.find(prop => prop.type === prop_type);
+    const config = await Config.getContactType(place_type);
```
freddieptf, line 37: Could be reverted, doesn't look relevant to the changes in the PR

src/routes/api.ts
```diff
@@ -0,0 +1,37 @@
...
+      const { config } = ConfigFactory.getConfigFactory().loadConfig();
+      reply.status(200);
+      return config;
+    } catch (error) {
+      reply.status(404).send({ message: 'No configuration found' });
+    }
+
+  });
+
+  fastify.post('/api/config', async (req: FastifyRequest, reply: FastifyReply) => {
```
freddieptf, line 25: I was able to post a config without any domains, which crashed the app after a logout. I was unable to post a new config after the crash happened

src/config/config-factory.ts
```diff
@@ -1,24 +1,107 @@
...
+      this.config = this.readConfig();
+      console.log(`Using uploaded configuration: ${this.filePath}`);
+      return this.config;
+    }
+
+    //set default config
+    if (!key) {
+      console.log('no configuration found. Setting to default');
+      const defaultConfig = kenyaConfig;
+      defaultConfig.config.domains = [{
```
freddieptf, line 49: Wouldn't this override the domains in this config? Would it be better to just crash the app if the key isn't set in the environment variables since that would be a deployment config issue?

freddieptf: The crash scenario is just below this code block which i think is a better way of handling this

src/config/config-factory.ts
```diff
@@ -1,24 +1,107 @@
...
+      return this.config;
+    }
+    return null;
+  }
+
+  writeConfig(jsonConfigData: ConfigSystem): void {
+    try {
+      const jsonString: string = JSON.stringify(jsonConfigData, null, 2);
+      fs.writeFileSync(this.filePath, jsonString);
+      this.refreshConfig();
```
freddieptf, line 82: since `refreshConfig` is only used here, maybe you can just directly set the config here after writing it. Saves you a trip to the filesystem. 

nitpick: the parameter name `jsonConfigData` is kinda confusing, maybe just `config` is fine?

src/routes/app.ts
```diff
@@ -12,7 +12,7 @@ import WarningSystem from '../warnings';
 
 export default async function sessionCache(fastify: FastifyInstance) {
   fastify.get('/', async (req, resp) => {
-    const contactTypes = Config.contactTypes();
+    const contactTypes = await Config.contactTypes();
```
freddieptf, line 15: No longer async

freddieptf (changes requested):

src/config/config-factory.ts
jonathanbataire: lets leave it there keeps it organized the writeconfig will be doing more than it describes 
 besides its a great action to have on the class

src/config/config-factory.ts
jonathanbataire: i have added default config in case env variable is not provided this can then be overridden via api

freddieptf: environment variables should always be provided and they are checked at start up. I cannot recommend changing env variables via API. We also shouldn't be changing the config values in code

jonathanbataire: thats not what im doing 
i simply making the env variable optional
if not provided the app falls back to a default config which allows overwriting via api with any auth domain
other configs ie env and uploaded will authenticate domain before overwriting 
chck it out

src/server.ts
jonathanbataire: resolved

src/config/index.ts
jonathanbataire: updated

src/routes/api.ts
jonathanbataire: resolved

src/config/index.ts
```diff
@@ -195,22 +203,45 @@ export class Config {
...
+      existingDomain => _.some(assertionConfig.domains, newDomain => _.isMatch(existingDomain, newDomain)));
+    if (!this.isDefaultConfig() && !newConfigHasExistingDomain) {
+      throw Error(`Invalid configuration: 'domains' property doesn't contains any existing domain`);
+    }
+
+    if (!assertionConfig.contact_types || assertionConfig.contact_types.length === 0) {
+      throw Error(`invalid configuration: 'contact_types' property is empty`);
+    }
+
+    for (const contactType of assertionConfig.contact_types) {
```
freddieptf, line 236: We need to update this and make validation very thorough making sure all required fields on the `ContactType` object have values. Use the object definition as a guide on what is strictly required. For eg. I can upload a `ContactType` without the `name` field even though it's required

src/server.ts
```diff
@@ -72,6 +72,17 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
   checkRedisConnection();
 
   fastify.addHook('preValidation', async (req: FastifyRequest, reply: FastifyReply) => {
+    if (req.routeOptions.url?.startsWith('/api') && req.method === 'POST') {
```
freddieptf, line 75: Shouldn't this be for all request methods?

freddieptf (changes requested):
Good progress, there are some changes around having a "default config" that i feel should not be a part of this PR and require further discourse. Please create another issue for this and lets have this PR only address adding the API

src/config/index.ts
```diff
@@ -195,22 +203,45 @@ export class Config {
...
 
+  public static isDefaultConfig(): boolean {
+    const { config } = ConfigFactory.getConfigFactory().loadConfig(CONFIG_NAME);
+    return config.domains.some(c => c.domain === 'default-configuration-domain');
+  }
+
   // TODO: Joi? Chai?
-  public static assertValid({ config }: PartnerConfig = partnerConfig) {
-    for (const contactType of config.contact_types) {
+  public static assertValid(config?: PartnerConfig) {
```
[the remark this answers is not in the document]
jonathanbataire: not sure i understand this fully

src/config/index.ts
```diff
@@ -195,22 +203,45 @@ export class Config {
...
+    const { config: assertionConfig } = config || ConfigFactory.getConfigFactory().loadConfig(CONFIG_NAME);
```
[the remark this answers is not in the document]
jonathanbataire: i did this cause the way it was built initially 
some in some cases assertValid is called with any config passed
so to minimized changes i had to carter for these scenarios
it not doing anything different from what was there only how its executed

src/config/index.ts
```diff
@@ -195,22 +203,45 @@ export class Config {
...
+
+    if (!assertionConfig.domains || assertionConfig.domains.length === 0) {
+      throw Error(`invalid configuration: 'domains' property is empty`);
+    }
+
+    const availableDomains = Config.getDomains();
+    const newConfigHasExistingDomain = _.some(availableDomains,
+      existingDomain => _.some(assertionConfig.domains, newDomain => _.isMatch(existingDomain, newDomain)));
+    if (!this.isDefaultConfig() && !newConfigHasExistingDomain) {
+      throw Error(`Invalid configuration: 'domains' property doesn't contains any existing domain`);
```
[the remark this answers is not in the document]
jonathanbataire: if you have domains a,b, and c in the tool
when updated you dont want to overwrite them with x,y and z without atleast having any of a,b and c in the new config
avoid locking yourself out accidentally/intentionally

src/config/index.ts
jonathanbataire: mainly focused on things that would crash the app

src/server.ts
jonathanbataire: this is checking for credentials not all requests have credentials
only api call come with credentials

src/config/default-config/config.json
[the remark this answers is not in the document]
jonathanbataire: if you leave ke config as dafault i cant update config in my set up if i dont have admin access to ke instance

src/config/index.ts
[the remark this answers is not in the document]
freddieptf: > some in some cases assertValid is called with any config passed

and that's exactly how it's supposed to work, to validate the config passed to the function. Please revert.

src/config/index.ts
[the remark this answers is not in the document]
freddieptf: You made the parameter `config` optional. It shouldn't be

src/config/index.ts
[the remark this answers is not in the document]
jonathanbataire: i dont understand before this PR this function is called without passing config into it 
its what im trying to maintain here

src/config/index.ts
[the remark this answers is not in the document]
freddieptf: What if you only had credentials for domain `d` and it gets overwritten leaving domains `a, b, x, y, z`? Wouldn't you still be locked out? 

That doesn't make the config invalid and i don't think it's something we should be checking for. Admins have access to the deployments *(cht and this tool)* and should be able to handle such cases so i don't see how that would be a problem

src/config/index.ts
freddieptf: Any required fields missing from `ContactType` will crash the app at some point. The app functionality i.e all generated forms, heavily depends on `ContactType`.

src/config/index.ts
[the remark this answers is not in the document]
jonathanbataire: dont you think a person with credentials to a domain not part of the UMT domains shouldnt be able to overwrite domains
scenario: if i had admin access to ug instance should i be able to use the same admin access to upload config in a ke instance

src/server.ts
freddieptf: Shouldn't the check you implemented apply for all api calls? Why limit it to `POST` calls only?

src/config/index.ts
jonathanbataire: i have tested this the app will only misbehave but wont crush unless there is a particular scenario im missing
if the app is misbehaving config can be update ideally it should be tested before deploying
at least for now 
lets handle these other cases in a new pr trying to get this in asap its blocking work

src/server.ts
jonathanbataire: so you can call this endpoint in a browser to view config

src/config/default-config/config.json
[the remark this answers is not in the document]
freddieptf: > not having it leaves a gap on how config upload workflow would go

Isn't the workflow simply overwriting the app's current config?

> if you leave ke(or other) config as default i cant update config in my set up if i dont have admin access to ke instance
> also remember config is not set only in env variable

The config to load is already set via the `CONFIG_NAME` env variable. If no part of this api touches the env variables then how would that value change? For your case wouldn't It always be `chis-ug`?

src/config/index.ts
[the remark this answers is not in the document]
freddieptf: It had a default value, that doesn't mean it was optional. I don't think you should be able to pass an `undefined` value to this function

src/config/default-config/config.json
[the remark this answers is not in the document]
jonathanbataire: built in config file is never overwritten
a new file is created for the uploaded config
NOTE as of last week chis-ug is no longer a built in config it was removed by some PR

src/config/index.ts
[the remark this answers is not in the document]
jonathanbataire: thats exaclty what it means :) 
check this out
https://github.com/medic/cht-user-management/blob/6547f539021b1ba02aed0ce8a859f979d341c7e8/src/index.ts#L10

src/config/index.ts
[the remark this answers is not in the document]
freddieptf: It had a default value that was the config loaded at startup. The equivalent behavior now would be calling `loadConfig` and passing that value to this function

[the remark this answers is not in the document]
jonathanbataire: similar to what is being discussed here https://github.com/medic/cht-user-management/pull/240#discussion_r1998899665

src/config/index.ts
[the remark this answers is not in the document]
freddieptf: This doesn't prevent that though. If that's what you're trying to prevent then shouldn't you be checking that the domain passed in the api call exists in the config's domain list?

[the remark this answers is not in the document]
jonathanbataire: let me explain better 
if i had admin access to ug i can overwrite all domains and change them to all ke domains 
this check tries to avoid this such that at least one of the new domains exited in the previous config
Overwriting all is NOT impossible this just makes it harder so its not done by accident 
Overwriting all would be at least a 2 step operation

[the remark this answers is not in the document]
jonathanbataire: hence avoid admin locking them selves out by uploading all ke domains in a ug app and locking themselves out because they dont have ke credentials

src/config/index.ts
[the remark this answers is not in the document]
freddieptf: It had a default value. The default value was set outside the function, you are setting the default value **inside** the function and changing the function signature to allow for a null value. That's my issue here. The function is meant to validate a config that's **passed** to it via the parameters.

src/config/index.ts
freddieptf: The app won't work correctly if a `ContactType` doesn't have a valid `name`. The data clean-up you'd set yourself up for just from this one field missing is worth having the checks in place now. This is just one example.

src/config/index.ts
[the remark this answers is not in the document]
jonathanbataire: also the default param value was being set in the function not outside

src/config/default-config/config.json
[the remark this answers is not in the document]
jonathanbataire: no env variable
deploy app
upload config from project repo

src/config/index.ts
jonathanbataire: i dont think cht would accept uploading data if the `name` is invalid/wrong
and for this is unlikely to happen given config has to go thru testing by the dev before deployment
again this wont break the app lets handle all these scenarios in another pr

src/config/index.ts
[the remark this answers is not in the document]
jonathanbataire: honestly i dont see any problem with that check its just a precaution
that scenario is handled

src/config/index.ts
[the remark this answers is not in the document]
jonathanbataire: let me know if you want to hope on a call tomorrow so we can finalize this

src/config/config-factory.ts
```diff
@@ -1,24 +1,105 @@
...
+  private config: PartnerConfig | null = null;
+  private static configFactoryInstance: ConfigFactory;
+  private filePath: string;
+
+  private DEFAULT_CONFIG_MAP: { [key: string]: PartnerConfig } = {
+    'CHIS-KE': kenyaConfig,
+    'CHIS-TG': togoConfig,
+    'CHIS-CIV': civConfig,
+    'CHIS-ML': maliConfig,
+    DEFAULT: defaultConfig
```
kennsippell, line 20: can you just add the config for uganda back into the repo? it should can a flag like `can upload` to true to ensure configs on other instances are not replaceable

src/config/config-factory.ts
```diff
@@ -1,24 +1,105 @@
...
+  };
+
+  constructor() {
+    this.filePath = path.join(this.getConfigUploadDirectory(), 'config.json');
+  }
+
+  static getConfigFactory(): ConfigFactory {
```
kennsippell, line 27: if this is a singleton, why not just have ConfigFactory be a static class?

src/config/config-factory.ts
```diff
@@ -1,24 +1,105 @@
...
+  private DEFAULT_CONFIG_MAP: { [key: string]: PartnerConfig } = {
+    'CHIS-KE': kenyaConfig,
+    'CHIS-TG': togoConfig,
+    'CHIS-CIV': civConfig,
+    'CHIS-ML': maliConfig,
+    DEFAULT: defaultConfig
+  };
+
+  constructor() {
+    this.filePath = path.join(this.getConfigUploadDirectory(), 'config.json');
```
kennsippell, line 24: `filePath` should be explicit about what file it is pointing at. maybe `uploadedConfigPath`

src/config/config-factory.ts
```diff
@@ -1,24 +1,105 @@
...
+    //set default config
+    if (!key) {
+      console.log('no configuration found. Setting to default');
+      const defaultConfig = this.DEFAULT_CONFIG_MAP.DEFAULT;
+      return this.config = defaultConfig;
+    }
+
+    const usingKey = key.toUpperCase();
+    console.log(`Using configuration: ${key}`);
+    const result = this.DEFAULT_CONFIG_MAP[usingKey];
```
kennsippell, line 56:
```suggestion
    this.config = this.DEFAULT_CONFIG_MAP[usingKey];
```

This would be cleaner without the intermittent variable `result` imo

src/config/config-factory.ts
```diff
@@ -1,24 +1,105 @@
...
+      fs.writeFileSync(this.filePath, jsonString);
+      this.refreshConfig({ config });
+    } catch (error) {
+      throw new Error('writeConfig: Failed to write file');
+    }
+  }
+
+  private getConfigUploadDirectory(): string {
+    const configDir = path.join(__dirname, '..', 'config_uploads');
+    if (!fs.existsSync(configDir)) {
```
kennsippell, line 88: better to have this existsSync check in `writeConfig`?

src/lib/authentication.ts
```diff
@@ -57,4 +58,16 @@ export default class Auth {
...
     return ChtSession.createFromDataString(data);
   }
+
+  public static async apiAuth(username: string, password: string, domain: string) {
+    const authInfo = Config.getAuthenticationInfo(domain);
+    try {
+      const chtSession = await ChtSession.create(authInfo, username, password);
+      return chtSession;
+    } catch (error: any) {
+      console.error(`Login error: ${error}`);
```
kennsippell, line 68: this is a bit of a weird function i think. 

1. why does it exist. it just calls ChtSession.create... i think it is largely obfuscating what is happening
2. why are we suppressing exceptions like this? and returning an empty object? weird
3.  logging that there was a "login error"?

can you re-evaluate what's going on and try to clean up?

src/lib/cht-session.ts
```diff
@@ -31,13 +31,15 @@ export default class ChtSession {
...
   public readonly chtCoreVersion: string;
+  public readonly isAdmin: boolean;
 
   private constructor(creationDetails: SessionCreationDetails) {
     this.authInfo = creationDetails.authInfo;
     this.username = creationDetails.username;
     this.facilityIds = creationDetails.facilityIds;
     this.sessionToken = creationDetails.sessionToken;
     this.chtCoreVersion = creationDetails.chtCoreVersion;
+    this.isAdmin = this.facilityIds.includes(ADMIN_FACILITY_ID);
```
kennsippell, line 42: dont love this change.  you're trading off space vs time complexity (caching) for something that doesnt really need to be cached... it appears to be called in very few locations.  can you explain your rationale?

src/routes/api.ts
```diff
@@ -0,0 +1,37 @@
+import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify';
+import ConfigFactory from '../config/config-factory';
+import { Config, ConfigSystem } from '../config';
+
+export default async function api(fastify: FastifyInstance) {
+  fastify.get('/api', (req: FastifyRequest, reply: FastifyReply) => {
```
kennsippell, line 6: can we remove this? is it adding value? would prefer to version endpoints like `/api/v1/config`

src/routes/api.ts
```diff
@@ -0,0 +1,37 @@
...
+    });
+  });
+
+  fastify.get('/api/config', async (req: FastifyRequest, reply: FastifyReply) => {
+    try {
+      const { config } = ConfigFactory.getConfigFactory().loadConfig();
+      reply.status(200);
+      return config;
+    } catch (error) {
+      reply.status(404).send({ message: 'No configuration found' });
```
kennsippell, line 20: i dont think 404 is the right response.  that would imply the `/api/config` endpoing wasn't found, but it was found... it just just returning empty.  i think 200 with empty response or message is better

src/routes/api.ts
kennsippell: gj finding this case but there will be others. i'd bet on it.
this points to a deeper problem - how do you recover in production if an invalid config crashes the app? just another reason this pattern is concerning

src/config/config-factory.ts
```diff
@@ -1,24 +1,105 @@
...
+      const jsonString: string = JSON.stringify(config, null, 2);
+      fs.writeFileSync(this.filePath, jsonString);
+      this.refreshConfig({ config });
+    } catch (error) {
+      throw new Error('writeConfig: Failed to write file');
+    }
+  }
+
+  private getConfigUploadDirectory(): string {
+    const configDir = path.join(__dirname, '..', 'config_uploads');
```
kennsippell, line 87: what sort of scenarios have you tested to ensure proper persistence of this file in docker?

src/routes/api.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+  });
+
+  fastify.post('/api/config', async (req: FastifyRequest, reply: FastifyReply) => {
+    try {
+      const config = await req.body as ConfigSystem;
+      Config.assertValid({ config });
+      ConfigFactory.getConfigFactory().writeConfig(config);
```
kennsippell, line 29: seems writeConfig should call `assertValid`? that feels like the right flow to ensure invalid configs cannot be written

src/server.ts
kennsippell: if this is specific to a single endpoing, why isn't it in the `route` file. why does it have to awkwardly be in this hook?

src/server.ts
```diff
@@ -80,6 +91,32 @@ const build = (opts: FastifyServerOptions): FastifyInstance => {
...
     }
 
+    const authHeader = req.headers.authorization as string;
+    if (authHeader && authHeader.startsWith('Basic ')) {
+      if (!req.routeOptions.url?.startsWith('/api')) {
+        reply.status(404).send({ error: 'not found' });
+        return;
+      }
+
+      const { domain } = req.query as { [key: string]: string };
```
kennsippell, line 101: is basic auth really what we want to use? why not just have an authenticated API? imo we should require users of the API to use `/login` (one way to authenticate to UMT), get issued a cookie, and then include it in requests to `/api`.  that seems straightforward and not terribly burdensome to users of the API (atm just jonathan)

test/config.spec.ts
```diff
@@ -1,18 +1,25 @@
 import { expect } from 'chai';
 
 import { Config, PartnerConfig } from '../src/config';
-import { CONFIG_MAP } from '../src/config/config-factory';
+import DEFAULT_CONFIG_MAP from '../src/config/config-factory';
 import { mockSimpleContactType } from './mocks';
```
kennsippell, line 5: add tests for `config-factory`?

kennsippell (changes requested):

Issue #185, opened 2024-06-27
Support for having multiple places assigned to a user


kennsippell:
CHT Core 4.9 supports having multiple places assigned to a single user
https://docs.communityhealthtoolkit.org/core/releases/4.9.0/#allow-multiple-places-to-be-assigned-to-users

In Kenya, we are planning to have a single CHA account with multiple CHUs associated with it.  How can we support this scenario and make CHU management easy?

ChinHairSaintClair:
Hi @kennsippell ,

Have you noticed any adverse replication impact with the "multi places assignment" feature? 
Or should one be a little more aggressive with the purge script? 
We've noticed that our online roles start running into timeouts when requesting around 28k records in the report view. 
Does the purge script also reduce the amount of data being pulled through to the user management tool?

kennsippell:
Hey @ChinHairSaintClair.  I don't think anybody is using the multiplace assignment feature yet. We are still getting ready.

In general, that change didn't changed the CHT's client-side performance characteristics except to make it easier to have too many docs associated with a single account.  When we use this feature, all the normal tricks to keep acceptable CHT performance will be even more relevant: purging (as you mention), but also replication depth since this feature is designed with supervisors in mind.

One clarification is that I think the feature is for offline-users only (maybe I'm wrong? I don't plan to use it for that users); but likely no impact for online users.  Purging has no impact on the amount of data pulled through the user management tool, but this tool only deals with contacts and never looks at reports.

ChinHairSaintClair:
Hi @kennsippell. That's fair enough, only got added in the latest release (4.9 at the time of writing).
We're pushing to upgrade and then intend to use this feature bit of an unorthodox way.
Operationally we need to track people that "out-migrate", and often the movements are outside the scope of our CHWs. Unfortunately at the same time, since they have intimate knowledge on the lay of the land, the CHWs are often best placed to indicate where exactly the person moved to. Bit of a predicament. Thus the `halfway house` concept is born.
The hope is to create a `halfway house` per `NPO` and assign each CHW to both their own area of operation, but then also to this item. It would enable each CHW across the system to move a patient out of a household, detailing in text where they have moved to physically, and then place them into this `halfway house` for further processing by the `site managers`.
The `site managers` , having elevated scope , will then be able, using the movement notes, to move the patient into an appropriate hierarchy item (should the item exist). 
We'll then also be able to identify areas where the app needs to start operating in.
Interested to know what you think of this idea, or have any concerns.

[...]

Issue #70, opened 2024-02-13
Support for creating supervisor users at a place that already exists


kennsippell:
For Uganda, there was a single bulk import of places at project inception.  Now they wish to bulk import supervisors and attach them to existing places.

kennsippell:
Still unclear about what to do when multiple supervisors are at the same place - is this a replacement or this is totally fine. Check out the data on ug-vht and let's see how things are set up.

PR #241, opened 2025-01-17
Script to import and reassign CHAs with multiple CHP Areas

kennsippell:
Script used for Digital Payment task https://app.clickup.com/t/1278154/CHT-13271 to create one account per CHA in Nairobi and Migori. 

[Here](https://docs.google.com/spreadsheets/d/1RptjqUQF1MC0O4utzfrYHJg-EBU3g6wiXsSZfnMBcQc/edit?gid=37999009#gid=37999009) is the input for Nairobi for once the instance is back up.

Progress toward #185
Progress toward #70

cc @eljhkrr as FYI


kennsippell:
@freddieptf This was used to create CHA users in Nairobi this week. Would you kindly review at your convenience?

kennsippell:
@freddieptf I've updated this scripts to also support CSV-based bulk reassignment of CHUs. It's quite rough.

The script does the following:
1. Accepts the standard minimal format for CHA replacement
2. Downloads all primary contacts for all CHUs
3. Attempts a string-match by primary contact's name and subcounty
4. If no match is found, it prompts the user if these CHA accounts shoheuld be created
5. If match is found, it "reassigns" the CHU to all user accounts associated with the primary contact's place

I considered three approaches for this broadly, focusing on the question "how do you know the username of the CHA who is the recipient of the CHU"

1.  Get the user to put the username directly into the CSV.  I cons I saw are that it required a different CSV format, and is a bit harder for users. But I maybe I wish I had in the end.
2. Lookup the user based on the "fullname" attribute in the users table. The cons I saw here are that it is difficult to group-by subcounty, there is no API endpoint to get users by role, the fullname of the user isn't always accurate or highly visible.
3. Do a fetch of all primary contacts of CHUs.  The cons of this is that it is expensive, and the result is ambiguous: if we are looking for the username of "Tom" and we find two of him and then look up usernames for those two places - there could be 10 users? I did this and updated all.

Imlementation notes:

* I've created this as a script because I'm unclear how well this work will align with the UI work you're leading https://github.com/medic/cht-user-management/issues/185
* Some aspects which seem very likely to be needed or unavoidable, I've built into UMT
* If you plan to use an approach like this, a crux will be getting the primary contact information for string comparison. We dont fetch it now and if we ever did this for CHP-level areas, it'd be slow.

freddieptf:
> Downloads all primary contacts for all CHUs
> Attempts a string-match by primary contact's name and subcounty

Is this necessary for re-assignments? if it's re-assigning then I'm assuming the CHA already exists in some CHU? Couldn't we just use the CHU which is unique and get the primary contact/user from there?

kennsippell:
> Is this necessary for re-assignments?

I gave three options for how to approach this in the text above.  Sounds like you are proposing a fourth option here? The inputs you are imagining are something like:

1. "For the CHA who manages X CHU"
2. "They should now also manage Y CHU"

As mentioned above, I took the approach I did because it allows for the same template that exists today.  If you have opinions about the best approach, I am all ears. Or I am available to discuss.

kennsippell:
CHA today had problems with their account because the contact_id on the user was deleted. Need to update this script to avoid that.

PR #246, opened 2025-02-04
hide additional place fields in the form when it's a replacement

freddieptf:
This makes it so the only fields that can be updated during the replacement (on the UI) are the primary contact, any edits on the place have to be done on the CHT instance instead. 

From discussion in https://github.com/moh-kenya/config-echis-2.0/issues/2651#issuecomment-2586329672


paulpascal (approved):
Straight 🍞

PR #250, opened 2025-02-06
feat(#243): prevent login into UMT without required permissions

### LINKED ISSUE #243: Do not allow users to login if they don't have the necessary permissions
Users need 5-6 permissions to use the user management tool.  If they don't have all the required permissions, they sould not be able to login.

Benmuiruri:
This PR introduces a check to prevent login into UMT if a user does not have the required permissions to use the features in UMT 

The `user-permission service` validates that the user's roles have the required permissions. 

<details>
<summary>Video showcase</summary>

https://github.com/user-attachments/assets/664aafaa-3ffc-4e24-986d-207ea66ac6a0

</details>

Closes #243


src/lib/cht-session.ts
```diff
@@ -11,6 +11,7 @@ import { RemotePlace } from './remote-place-cache';
 const COUCH_AUTH_COOKIE_NAME = 'AuthSession=';
 const ADMIN_FACILITY_ID = '*';
 const ADMIN_ROLES = ['admin', '_admin'];
+const AUTHORIZED_ROLES = [...ADMIN_ROLES, 'user_manager'];
```
freddieptf, line 14: The `user_manager` role is a part of cht app settings and thus configurable. Maybe we can make this dynamic by loading the configured role from the UMT config?

src/lib/cht-session.ts
```diff
@@ -95,6 +96,14 @@ export default class ChtSession {
     return setCookieHeader?.[0]?.split(';')
       .find((header: string) => header.startsWith(COUCH_AUTH_COOKIE_NAME));
   }
+
+  private static validateUserRole(userDoc: any, username: string): void {
+    const hasAuthorizedRole = (roles: string[]): boolean => _.intersection(AUTHORIZED_ROLES, roles).length > 0;
+
+    if (!hasAuthorizedRole(userDoc?.roles || [])) {
+      throw Error(`User ${username} role does not have the required permissions`);
```
freddieptf, line 104: What happens when the user has the role but the role isn't configured with the right permissions?

Benmuiruri: Here is my understanding. The role to perform actions on UMT is configured first then a user who can interact with UMT for instance `user_manager` is created with that role that has the permissions.

So a user who has role`['county_supervisor', 'user_manager']`. Such a user will be able to login to UMT because we will have added the check for `user_manager`

A user who only has `['county_supervisor'] `will not be able to login because they don't have the role with the permissions. 

Is your scenario still a concern from this perspective ?

src/lib/cht-session.ts
Benmuiruri: Thanks for the idea. When you say UMT config, which file is that I take a look ?

freddieptf: eg https://github.com/medic/cht-user-management/blob/main/src/config/chis-ke/config.json

src/lib/cht-session.ts
freddieptf: Yes because a cht admin could have configured the `user_manager` role without a permission like "can_create_users" or something like that. So even if they have the role, they don't have the required permissions. I'm not sure if it's possible to check if the role/user has the required permissions but ideally that would atleast strictly ensure one can't login. If not possible then this is fine

Benmuiruri: I changed the approach. Now each config in the UMT will define which are the required permissions users need to be allowed access. 

The session now checks whether the user's roles have the permissions. The question now is, which permissions do we enforce for Echis ke ?

freddieptf: They are defined in https://github.com/moh-kenya/config-echis-2.0/blob/9a140948070d0191990049dfb1fa9433165a99bb/app_settings/base_settings.json#L146

I don't think they need to be part of the uMT config as we already know beforehand what permissions are needed

Benmuiruri: I've moved the permissions list to the service. The list of permissions needed to log into UMT will be defined in the service.

src/lib/cht-session.ts
kennsippell: Or ask the CHT for the user's permissions and see if they have the permissions that are actually required?

src/routes/authentication.ts
```diff
@@ -36,9 +36,11 @@ export default async function authentication(fastify: FastifyInstance) {
       chtSession = await ChtSession.create(authInfo, username, password);
     } catch (e: any) {
       console.error(`Login error: ${e}`);
+      const errorMessage = e.message?.includes('permissions') ? e.message : 'Login Failed. Please Try Again';
```
kennsippell, line 39: Is it always safe to display the message? Would be nice to at least test those scenarios.   If the message is safe, maybe we should just always display it? If the message isn't safe, then this probably isn't a sufficient check ... like what if the username includes `permissions` and is in message?

Benmuiruri: Let me revert to how it was.  The better way would be creating an explicit Error class but that's beyond the scope of this task.

 Let's communicate vaguely to end users with login failed as before. Devs will be able to check the `e.message` safely.

kennsippell: https://github.com/medic/cht-user-management/issues/253

src/lib/authentication-error.ts
```diff
@@ -33,4 +33,8 @@ export class AuthError extends Error {
   static INSTANCE_OFFLINE() {
     return new AuthError(`Unable to connect to instance. Please check instance availability.`);
   }
+
+  static MISSING_PERMISSIONS(username: string) {
+    return new AuthError(`User ${username} role does not have the required permissions`);
```
freddieptf, line 38: Maybe we should drop the word role here?

```suggestion
    return new AuthError(`User ${username} does not have the required permissions`);
```

freddieptf (approved):
LGTM

PR #252, opened 2025-02-10
fix(#236): update titleCase to handle hyphens and parenthesis

### LINKED ISSUE #160: Title Casing after hyphen
`Eor-Ekule CHU` is uploaded as `Eor-ekule Community Health Unit` but should be `Eor-Ekule Community Health Unit`

### LINKED ISSUE #236: Improper casing for names where part is in brackets
Observed: `Matopeni B (health Express)`
Expected: `Matopeni B (Health Express)`

Benmuiruri:
Update `titleCase` function to handle hyphens and parenthesis. 

Closes #236 
Closes #160


src/validation/validator-name.ts
```diff
@@ -36,12 +36,34 @@ export default class ValidatorName implements IValidator {
...
+    }
+
+    const titleCase = (word: string) => word[0]?.toUpperCase() + word.slice(1);
+    const isRomanNumeral = /^[ivx]+$/i;
+
+    const processWord = (word: string) => {
+      if (word.match(isRomanNumeral)) {
+        return word.toUpperCase();
+      }
+      return word.includes('-')
```
kennsippell, line 50: here we are splitting on a symbol, calling `processWord` for each word, and then rejoining. we are doing this again on line 62-65.  

what do you think of putting that pattern into a function `splitAndProcess(value, ' ')` and `splitAndProcess(value, '-')`?

src/validation/validator-name.ts
```diff
@@ -36,12 +36,34 @@ export default class ValidatorName implements IValidator {
...
-    const words = value.toLowerCase().split(' ');
-    const titleCase = (word: string) => word[0].toUpperCase() + word.slice(1);
-    const isRomanNumeral = /^[ivx]+$/ig;
-    const titleCased = words
+    if (!value) {
+      return '';
+    }
+
+    const titleCase = (word: string) => word[0]?.toUpperCase() + word.slice(1);
+    const isRomanNumeral = /^[ivx]+$/i;
```
kennsippell, line 44: its clear and useful to have the regex captured via a descriptive name.  it would be nice to do the same for the regex on lines 56-58.  in general, i find regex hard to parse and read and benefit from either a description or a comment

kennsippell (approved):

src/validation/validator-name.ts
Benmuiruri: Good catch.

PR #254, opened 2025-02-11
chore(#253): add descriptive login error messages

### LINKED ISSUE #253: Descriptive messages as-to why login fails
* Incorrect username/password
* Instance down
* Insufficient permission

Benmuiruri:
Makes login error messages more descriptive. 
 
 Closes #253


Benmuiruri:
Hi @kennsippell  I thought of different ways to design the solution and came up with this. Please share your thoughts of it.

src/lib/authentication/authentication.ts
kennsippell, line 1: why is the folder required if there is only one file in it? if there is no explicit plan to have more files in here, i suspect we should not create the folder?

src/routes/authentication.ts
```diff
@@ -30,28 +38,36 @@ export default async function authentication(fastify: FastifyInstance) {
...
-    const expires = Auth.cookieExpiry();
-    resp.setCookie(Auth.AUTH_COOKIE_NAME, tokenizedSession, {
-      signed: false,
-      httpOnly: true,
-      expires,
-      secure: true
-    });
+      resp.header('HX-Redirect', '/');
+    } catch (e: any ) {
+      if (e instanceof AuthError) {
```
kennsippell, line 60: suggest a new function `getLoginErrorMessage` which returns the error message? one call to render

src/lib/cht-session.ts
```diff
@@ -51,15 +52,21 @@ export default class ChtSession {
...
+      if (!sessionToken) {
+        throw new Error(`failed to obtain token for ${username} at ${authInfo.domain}`);
+      }
 
-    if (!sessionToken) {
-      throw new Error(`failed to obtain token for ${username} at ${authInfo.domain}`);
+      const creationDetails = await ChtSession.fetchCreationDetails(authInfo, username, sessionToken);
+      return new ChtSession(creationDetails);
+    } catch (e: any) {
+      if (e.response?.status === 401) {
```
kennsippell, line 65: Might be nice to have `AuthErrors.assertFromException(e)` for encapsulation?
I suspect the try/catch should be in createSessionToken? But that's up to you

src/lib/cht-session.ts
```diff
@@ -51,15 +52,21 @@ export default class ChtSession {
     return this.facilityIds.includes(ADMIN_FACILITY_ID);
   }
 
-  public static async create(authInfo: AuthenticationInfo, username : string, password: string): Promise<ChtSession> {
-    const sessionToken = await ChtSession.createSessionToken(authInfo, username, password);
+  public static async create(authInfo: AuthenticationInfo, username: string, password: string): Promise<ChtSession> {
+    try {
+      const sessionToken = await ChtSession.createSessionToken(authInfo, username, password);
+      if (!sessionToken) {
+        throw new Error(`failed to obtain token for ${username} at ${authInfo.domain}`);
```
kennsippell, line 59: Should this be an `AuthErrors`?

src/lib/cht-session.ts
```diff
@@ -51,15 +52,21 @@ export default class ChtSession {
...
+    } catch (e: any) {
+      if (e.response?.status === 401) {
+        throw AuthErrors.INVALID_CREDENTIALS();
+      }
+      throw e;
     }
-    
-    const creationDetails = await ChtSession.fetchCreationDetails(authInfo, username, sessionToken);
-    return new ChtSession(creationDetails);
   }
```
kennsippell, line 71: Can you check this out? AuthError? `throw Error(User ${username} does not have a facility_id connected to their user doc);`

and `ChtSession.assertCoreVersion`?

kennsippell (changes requested):
Nice!

kennsippell:
Tested this today with uasingishu instance. The login fails due to cert errors. I'm seeing the response hang and no error message ever appear.

src/routes/authentication.ts
```diff
@@ -30,28 +38,37 @@ export default async function authentication(fastify: FastifyInstance) {
...
 
-    const authInfo = Config.getAuthenticationInfo(domain);
-    let chtSession;
+    const getLoginErrorMessage = (error: unknown): string => {
+      if (error instanceof AuthError) {
+        console.error('Login error:', error.message);
+        return error.errorMessage;
+      }
+
+      console.error('Login error:', error);
```
kennsippell, line 47: can we log once after the call to `getLoginErrorMessage` to avoid multiple log statements?

kennsippell (approved):
Lookin' good!

src/lib/authentication-error.ts
```diff
@@ -0,0 +1,41 @@
+export class AuthError extends Error {
+  constructor(
+      public status: number,
```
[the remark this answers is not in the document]
Benmuiruri: This is outdated. I realized that its not always 401. Also we dont have access to `e` everywhere to access the statusCode. So I deleted it and just left the message. The `e` will be available in the logs together with the statusCode

src/lib/authentication-error.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  static INVALID_CREDENTIALS() {
+    return new AuthError('Invalid username or password');
+  }
+
+  static MISSING_CREDENTIALS() {
+    return new AuthError('Missing username or password');
+  }
+
+  static TOKEN_CREATION_FAILED(username: string, domain: string) {
```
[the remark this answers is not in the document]
Benmuiruri: Sure. 

I'm thinking we could use it in `createSessionToken` by checking if `(e.code === 'ECONNREFUSED')` ? would that be sufficient to know the instance is offline ?

Benmuiruri:
Hi @kennsippell could we get this over the line. Thanks

PR #257, opened 2025-02-17
Add form for assigning multiple places to a single user

### LINKED ISSUE #259: Add form for assigning multiple places to a single user
Part of #185 

Add a form for assigning multiple places to a single user for instances on cht >=4.9.0

freddieptf:
Part of #185 
Fixes #259 

This adds support for creating a new user with multiple new places at once.

Dependent on cht-core >= 4.9.0 and setting `can_assign_multiple` on a contact type in your config.json


src/lib/cht-api.ts
```diff
@@ -129,7 +129,7 @@ export class ChtApi {
   }
 
   async createUser(user: UserPayload): Promise<void> {
-    const url = `api/v1/users`;
+    const url = `api/v3/users`;
```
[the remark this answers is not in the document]
freddieptf: The new form only loads if the instance you logged in to is >=4.9.0 so i could create a `createUserV3` function  used only when the instance > 4.7 for backwards compatibility

src/liquid/new/new_place_btn.liquid
```diff
@@ -0,0 +1,9 @@
+<button
+  type="button"
+  hx-target="this"
+  hx-get="/place_form?place_type={{place_type}}"
+  hx-swap="outerHTML"
+  class="button is-small is-info mb-4"
+>
+  + New
```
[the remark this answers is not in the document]
freddieptf: `save and continue` lets the user add CHAs with a new list of places without being redirect to the homepage

src/liquid/new/place_list.liquid
```diff
@@ -0,0 +1,72 @@
+<div
+  class="my-4"
+  style="border: 1px solid #d3d3d3; border-radius: 4px; width: 100%;"
+  hx-get="/new/table?contact={{places[0].contact.id}}"
```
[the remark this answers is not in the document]
freddieptf: It's meant to help the user visualize the groups during the process of adding the places. It's not on the homepage because i felt it'd break the continuity of the list there

src/liquid/new/form.liquid
```diff
@@ -0,0 +1,60 @@
...
+    hx-swap="outerHTML"
+    hx-post="/new?place_type={{contactType.name}}"
+    style="width: 38rem;"
+  >
+    <div>
+      <h3 class="is-size-4">New {{ contactType.contact_friendly }}</h3>
+    </div>
+    <input name="place_type" value="{{contactType.name}}" hidden>
+
+    <p>Contact details</p>
```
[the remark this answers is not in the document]
freddieptf: > If I create a CHA and then try to add more places to it

You're meant to create a CHA with all their places using the "+New" button. When you click "save and continue" you save the CHA with their places and restart the process without navigating to the homepage

> Is there a way to add another CHU to a CHA that was created?

No there isn't, i'll add it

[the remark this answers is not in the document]
freddieptf: > Is there a way to add another CHU to a CHA that was created?

Added

src/liquid/new/form.liquid
```diff
@@ -0,0 +1,60 @@
+<div id="create_form_parent" class="is-flex is-flex-direction-column is-align-items-center">
```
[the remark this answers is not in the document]
freddieptf: Do you mean the top level navigation? Do we really need it there? All i included was the logo which takes you to the homepage so there's less distractions

src/lib/cht-api.ts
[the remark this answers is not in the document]
freddieptf: done

src/liquid/new/place_list.liquid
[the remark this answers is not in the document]
freddieptf: I've added the table to the homepage. It will show for any contact type has `can_assign_multiple` set to true

freddieptf:
@kennsippell I've add the new table to the homepage, it also includes links to pages for editing the contact and adding more places to the contact. 
Screenshot 2025-03-12 at 01-28-32 CHT User Management Tool

src/services/session-cache.ts
```diff
@@ -48,6 +49,28 @@ export default class SessionCache {
       .filter(p => options?.created === undefined || !!p.isCreated === options.created);
   };
 
+  public getPlacesForDisplay = (options?: {
+    type?: string;
```
kennsippell, line 53: can you make this a reusable type shared by `getPlacesForDisplay` and `getPlaces`?

src/services/session-cache.ts
```diff
@@ -48,6 +49,28 @@ export default class SessionCache {
       .filter(p => options?.created === undefined || !!p.isCreated === options.created);
   };
 
+  public getPlacesForDisplay = (options?: {
```
kennsippell, line 52: should this be called something like `getPlacesByContacts`

src/services/session-cache.ts
```diff
@@ -48,6 +49,28 @@ export default class SessionCache {
...
+    created?: boolean;
+    id?: string;
+    nameExact?: string;
+    nameIncludes?: string;
+  }): { canEdit?: boolean; places: Place[] }[] => {
+    const places =  this.getPlaces(options);
+    return Object.values(
+      _.groupBy(places, (p) => p.contact.id)
+    )
+      .reverse()
```
kennsippell, line 65: why the reverse?

src/services/session-cache.ts
```diff
@@ -48,6 +49,28 @@ export default class SessionCache {
...
+      .map(places => {
+        return {
+          places,
+          canEdit: !places.find(p => p.creationDetails.username)
```
kennsippell, line 69:
```suggestion
          canEdit: !places.some(p => p.creationDetails.username)
```

kennsippell, line 69:
```suggestion
          canEdit: !places.some(p => p.creationDetails.username)
```

src/services/session-cache.ts
```diff
@@ -48,6 +49,28 @@ export default class SessionCache {
...
+    type?: string;
+    state?: PlaceUploadState;
+    filter?: DirectiveFilter;
+    created?: boolean;
+    id?: string;
+    nameExact?: string;
+    nameIncludes?: string;
+  }): { canEdit?: boolean; places: Place[] }[] => {
+    const places =  this.getPlaces(options);
+    return Object.values(
```
kennsippell, line 62: i think this line would be more readable as multiple statements

src/services/upload-manager.ts
```diff
@@ -100,7 +176,7 @@ function getErrorDetails(err: any) {
   }
 
   if (err.response?.data?.error) {
-    return JSON.stringify(err.response.data.error);
+    return err.response.data.error.message ?? JSON.stringify(err.response.data.error);
```
kennsippell, line 179: if there is an error, return the message?  i'm unclear no this flow.

src/services/upload-manager.ts
```diff
@@ -19,13 +20,32 @@ export interface Uploader {
...
+    const independants = placesNeedingUpload.filter(p => !p.isDependant && !p.hasSharedUser);
+    const dependants = placesNeedingUpload.filter(p => p.isDependant && !p.hasSharedUser);
+
     await this.uploadPlacesInBatches(independants, chtApi);
     await this.uploadPlacesInBatches(dependants, chtApi);
+    await this.uploadGrouped(validPlaces.filter(p => p.hasSharedUser), chtApi);
+  };
+
+  uploadGrouped =  async (places: Place[], api: ChtApi) => {
+    const grouped = _.groupBy(places, place => place.contact.id);
```
kennsippell, line 38: should this use `getPlacesByContacts` interface?

kennsippell, line 38: should this use `getPlacesByContacts` interface?

src/services/upload-manager.ts
```diff
@@ -19,13 +20,32 @@ export interface Uploader {
 
 export class UploadManager extends EventEmitter {
   doUpload = async (places: Place[], chtApi: ChtApi, ignoreWarnings: boolean = false) => {
-    const placesNeedingUpload = places.filter(p => !p.isCreated && !p.hasValidationErrors && (ignoreWarnings || !p.warnings.length));
+    const validPlaces = places.filter(p => {
+      return !p.hasValidationErrors && (ignoreWarnings || !p.warnings.length);
```
kennsippell, line 24:
```suggestion
      !p.hasValidationErrors && (ignoreWarnings || !p.warnings.length)
```

src/services/upload-manager.ts
```diff
@@ -19,13 +20,32 @@ export interface Uploader {
...
+    });
+    const placesNeedingUpload = validPlaces.filter(p => !p.isCreated);
     this.eventedPlaceStateChange(placesNeedingUpload, PlaceUploadState.SCHEDULED);
 
-    const independants = placesNeedingUpload.filter(p => !p.isDependant);
-    const dependants = placesNeedingUpload.filter(p => p.isDependant);
+    const independants = placesNeedingUpload.filter(p => !p.isDependant && !p.hasSharedUser);
+    const dependants = placesNeedingUpload.filter(p => p.isDependant && !p.hasSharedUser);
```
kennsippell, line 30: can we make a third variable here `const grouped`? or alike?

src/services/upload-manager.ts
```diff
@@ -79,8 +99,64 @@ export class UploadManager extends EventEmitter {
...
+    for (const place of places) {
+      if (place.creationDetails.placeId) {
+        placeIds[place.id] = undefined;
+        continue;
+      }
+      this.eventedPlaceStateChange(place, PlaceUploadState.IN_PROGRESS);
+
+      try {
+        const payload = place.asChtPayload(api.chtSession.username, creationDetails.contactId);
+        await Config.mutate(payload, api, place.isReplacement);
```
kennsippell, line 116: i'm surprised this can't be implemented with more reuse. is a variation of `uploadSinglePlace` not possible?

src/liquid/new/form.liquid
```diff
@@ -0,0 +1,78 @@
...
+            hierarchy: hierarchy,
+            prefix: 'hierarchy_',
+            data: data,
+            disabled: show_place_form == true
+          %}
+        {% endif %}
+      {% endfor %}
+      {% for prop in contactType.contact_properties %}
+        {%
+          include "components/contact_type_property.html" prefix="contact_" prop=prop
```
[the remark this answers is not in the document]
freddieptf: `include` is deprecated, we should be using `render` https://shopify.dev/docs/api/liquid/tags/render

> When you render a snippet using the render tag, you can't use the [include tag](https://shopify.dev/docs/api/liquid/tags/include) inside the snippet.

There's this note though, which is why you might me use one or the other as it's not a direct drop in replacement. I think it can be handled when changing the templates to `.liquid` extension as it'd require some testing

src/routes/add-place.ts
```diff
@@ -132,27 +139,38 @@ export default async function addPlace(fastify: FastifyInstance) {
...
+    if (place.hasSharedUser) {
+      places = sessionCache.getPlaces({ type: place.type.name }).filter(p => p.contact.id === place.contact.id);
+    } else {
+      places = [place];
+    }
 
-    const chtApi = new ChtApi(req.chtSession);
     const uploadManager: UploadManager = fastify.uploadManager;
-    uploadManager.doUpload([place], chtApi, true);
+    uploadManager.doUpload(places, chtApi, true);
```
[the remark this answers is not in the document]
freddieptf: > so if user uploads one row, they upload all rows for the user? 

Yes. It's a compromise since the places all share one user and contact. If i allow uploading one by one then there's no guarantee the places will be created with the same user and contact.

> and this ignores warnings.

Warnings are ignored if you try upload a place directly from the list. That was the default behavior and it's carried over here.

src/services/upload-manager.ts
freddieptf: I was also not sure why error was being stringified so i assumed it had to be some versions of CHT returning different error objects. In my tests with CHT 4.16.0, the error object had a message field so i check for that before falling back to the previous implementation

src/services/upload-manager.ts
freddieptf: I don't think so, `doUpload` already gets the whole place list so all we have to do is filter and group

src/liquid/new/place_form_fragment.liquid
```diff
@@ -0,0 +1,41 @@
...
+        data=data
+        place_type=contactType.name
+      %}
+      {% if errors[prop.property_name] %}
+        <p class="mb-2" style="color: red;">{{ errors[prop.property_name] }}</p>
+      {% endif %}
+    {% endfor %}
+  </div>
+  <div class="is-flex is-justify-content-end is-align-items-center">
+    <span class="is-size-7 has-text-weight-semibold">Don't forget to click add before saving</span>
```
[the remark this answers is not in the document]
freddieptf: Not right now. Since there's no store for the partial place data that's on the form, there's no way to sanely handle validation errors and maintain the previous added places on the form

src/services/upload-manager.ts
freddieptf: There's not reuse of this function, it's only used once here

Issue #262, opened 2025-02-25
Due Mon Mar 3rd:  MoH Mali CHW User Managament Tool 


paulpascal:
**Partner Name**: MoH Mali CHW
**Proposed URL**: users-chis-ml.app.medicmobile.org
**Config Name**: users-chis-ml

**Action Items**
- [x]  verify URL: https://users-chis-ml.app.medicmobile.org
- [x]  ensure repo has [config](https://github.com/medic/cht-user-management/tree/main/src/config) for `users-chis-ml`. Directory and config name needs to match URL.  Should point to CHT Core production instance at `disc-mali.ml`
- [x]  create DNS entry: `CNAME` of `users-chis-ml.app.medicmobile.org` -> `k8s-prodchtalb-dcc00345ac-1792311525.eu-west-2.elb.amazonaws.com`
- [x]  create helm values file in [correct location](https://github.com/medic/cht-user-management/tree/main/scripts/deploy/values)
- [x]  `helm install` chart so the instance is live at above URL [per docs](https://github.com/medic/cht-user-management/blob/main/scripts/deploy/medic-deploy.md)
- [x]  update docs to have new instance in the `Known configurations` section

inromualdo:
@mrjones-plip, regarding the second request, @loukmantidjani will send them an email, and we will get back to you once we receive their response

cc: @paulpascal @hiellavis

paulpascal:
Thank you @mrjones-plip  for jumping on this!

1. Yes, there is a deadline for this request—next Monday.
2. I have discussed this with @hiell and @ro, and it seems that self-hosting is not yet in their pipeline, so there is no plan for that at the moment.

cc @hiell if you would like to add anything.

mrjones-plip:
DNS entry created in [Route 53](https://us-east-1.console.aws.amazon.com/route53/v2/hostedzones?region=us-east-1#ListRecordSets/Z3304WUAJTCM7P) and tested:

```
$ dig users-chis-ml.app.medicmobile.org +short
k8s-prodchtalb-dcc00345ac-1792311525.eu-west-2.elb.amazonaws.com.
18.169.26.92
13.42.201.181
```

![Image](https://github.com/user-attachments/assets/ebb8730a-d2f6-43f2-86b1-034223253f7a)

mrjones-plip:
[...]

PR #265, opened 2025-02-28
feat(#292): add moh mali chw to CD worfklow

paulpascal:
#262 Part 2

Here are adding the config to the CD workflow.


mrjones-plip (approved):
looks good!

PR #267, opened 2025-03-03
Update medic-deploy.md to add CD steps for new deployment

mrjones-plip:
This PR adds steps to add YAML to the CD script so that future deployments know how to get auto updated


paulpascal (approved):
Nice !

PR #268, opened 2025-03-03
chore: update CI config to run Build/Tests for PRs from forks

jkuester:


jkuester:
No pressure here, but I definitely plan to take what we decide here and try to implement the same pattern in the other repos (cht-core, cht-conf, etc).  :sweat_smile:

.github/workflows/run-tests.yml
```diff
@@ -1,10 +1,11 @@
 name: Run Build and Tests
 
-on: push
+on: [push, pull_request]
 
 jobs:
   test:
     name: Run Build and Tests
+    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository }}
```
kennsippell, line 8: if i didn't read your writeup, i wouldn't know what was up here...  worth a comment? maybe reference your PR write-up here?

.github/workflows/run-tests.yml
```diff
@@ -1,10 +1,11 @@
 name: Run Build and Tests
 
-on: push
+on: [push, pull_request]
```
kennsippell, line 3: > This would mean that the CI would no longer run for branches within the repo that are not associated with any pull request. On one had that is a "regression" in functionality

i guess i'd only call it a regression if we're affected _desired_ behavior. do we actually want to tests for commits w/o a PR for a concrete reason? what's the user scenario

kennsippell (approved):
> I promise that there is nothing to find but sadness and disappointment

omg lol.  You've been on a journey, my friend. you've seen things.   i'm approving cause i'm happy with what you've done, but a few comments below for discussion

Issue #149, opened 2024-04-18
Improved validation error text


kennsippell:
`Cannot find 'c_community_health_unit' matching Nakaale under Turkana South ` should be `Cannot find 'community health unit' matching Nakaale under Turkana South`

`Not a valid phone number for country code KE` sholud be `Not a valid phone number for country Kenya`

Error details should be shown on the edit page
Investigate as-to whether users find tooltip confusing and mitigate with UI hints as needed

Benmuiruri:
Hi @alexosugo this is not yet completed. I separated it from #272

alexosugo:
My bad.

PR #277, opened 2025-03-13
chore(#272): show missing feedback if login fails 

### LINKED ISSUE #272: When logging into eCHIS instances, login fails without any error or message
Currently, the nairobi instance is down. If I try to login, I see no error or feedback of any kind

https://github.com/user-attachments/assets/7fc8e69b-d973-4687-a07e-af5aa16664ee

Benmuiruri:
This PR: 
- Updates login error messages to show when instance is unreachable 
- Adds a loading state to the sign in button to improve user feedback after clicking `Sign In`

<details>
<summary>Unreachable instance error</summary>

https://github.com/user-attachments/assets/f20679ba-df5b-4f9e-adb8-01685dd3d76d

</details>

- Updates validation error text for KE phone numbers and unknown CU names

Closes #272

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -52,7 +52,6 @@ export type ContactProperty = {
   type: ContactPropertyType;
   required: boolean;
   parameter? : string | string[] | object;
-  friendly_country_name?: string;
   errorDescription? : string;
   unique?: string;
 };
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -52,6 +52,7 @@ export type ContactProperty = {
   type: ContactPropertyType;
   required: boolean;
   parameter? : string | string[] | object;
+  friendly_country_name?: string;
   errorDescription? : string;
   unique?: string;
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "cht-user-management",
-  "version": "2.1.9",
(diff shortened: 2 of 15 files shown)


src/config/index.ts
```diff
@@ -51,6 +51,7 @@ export type ContactProperty = {
   type: ContactPropertyType;
   required: boolean;
   parameter? : string | string[] | object;
+  friendly_country_name?: string;
```
kennsippell, line 54: can we keep config simple by using `not a valid KE phone number` instead of Kenya?

Benmuiruri: Sure... I added that because the issue description suggested Kenya instead of KE. I'll revert to use KE

Benmuiruri:
Hi @paulpascal this PR is ready for review when you have some time. Thanks.

src/config/index.ts
kennsippell: Sorry to give conflicting suggestions. I see that in the issue now. I personally doubt it is worth the complexity in config.  Does libphonenumber-js know that KE means Kenya?

> Cannot find 'c_community_health_unit' matching Nakaale under Turkana South  should be Cannot find 'community health unit' matching Nakaale under Turkana South

If this isn't fixed, I'd recommend not yet closing #149 with this PR.

paulpascal (approved):
Nice work @Benmuiruri

Issue #273, opened 2025-03-12
Expose prometheus endpoing for bullmq metrics


kennsippell:
Linked to work from #68

PR #278, opened 2025-03-17
Single endpoint for all prometheus metrics

kennsippell:
The fastify-metrics library doesn't play particularly nicely for extensibility. I attempted a few vectors to merge these two endpoints:

1. Interupt the response from fastify-metrics and append additional metrics from BullMQ
2. Obtain the `promClient` object and add the bullmq metrics as custom counters within fastify-metrics
3. Registry the fastify-metrics as its own endpoint and have /metrics invoke it

In the end, I liked the code for 1 the best. Let me know if you'd like to see some of the other approaches!

#273


mrjones-plip (changes requested):
While the code looks good as far as I can tell, I cannot get the `/metrics` endpiont to work - I get a `302` to login.  I actually went back to a branch [before](https://github.com/medic/cht-user-management/pull/267) you made your bullq change and the `/metrics` endpoint didn't work then either. I had to go all the way back to 4dafb28a8a33df944a868829a4a3737995aeb5dc - and then it finally worked!  Something changed between now and then - I can't see what off hand :thinking: 

However, researching what we did back then, I found the `watchdog-config` [directory](https://github.com/medic/cht-user-management/tree/main/scripts/deploy/watchdog-config)! This has all the files we just [recreated](https://github.com/medic/cht-watchdog/pull/146) in watchdog.    We should pick one winner (watchdog or UMT) and remove the watchdog config files from the loser.

I was involved with this earlier PR - sorry I didn't remember about these files already existing!

mrjones-plip:
Ah - sweet - thanks Kenn!  Nuking the `dist` folder did the trick.  Faster for me to run `npm run dev`, but I'm sure `docker-local-setup.sh` will work too.  `/metrics` works now - yay!

and yeah - go ahead and nuke the watchdog files here in favor of watchdog repo.  Maybe replace the [readme](https://github.com/medic/cht-user-management/blob/main/scripts/deploy/watchdog-config/readme.md) with a big "see watchdog repo" and a link?

mrjones-plip (approved):

PR #279, opened 2025-03-17
fix: dev section docs more opinionated, explicit call to `build`

mrjones-plip:
Quickie PR to redo the developer section to not give two choices and stick to one happy path which requires you to call `build`.  the prior version made it so `npm run build` was optional when it is very much not.


kennsippell (approved):
Smooth

PR #283, opened 2025-03-28
fix: removes can configure permission

jonathanbataire:
## Description
`can_configure` permission poses a security risk if user_managers log into CHT


kennsippell (approved):
I don't see how this causes a security risk. But I do agree that `can_configure` is not required to use UMT.

PR #288, opened 2025-04-15
fix(#269): hide navbar options on nested pages 

### LINKED ISSUE #269: UI Tophat allows uers to upload and save credentials when they aren't on homepage
A user should only have the option to upload or save credentials when they are on the homepage. It doesn't make sense to provide these options to users when they are in experiences like "New CHP" or "Move"

freddieptf:
Fixes #269

This hides the `Save Credentials` and `Upload` options in the drop-down on the top right for pages that aren't the home page


freddieptf:
The navbar component needs the contact type list to render the options in the `Add` and `Manage` menus

Issue #280, opened 2025-03-18
Rename all template extensions to `.liquid`


freddieptf:
1. Need to change all template extension to `.liquid` https://github.com/medic/cht-user-management/pull/257#discussion_r1999066372
2. Replace usage of `include` (deprecated) with `render` https://github.com/medic/cht-user-management/pull/257#discussion_r2000997120

Issue #270, opened 2025-03-06
Introduce Typed View Models


kennsippell:
The data passed to Liquid views is not currently typed via typescript. Our codebase can be improved if the data was typed.

PR #290, opened 2025-04-29
chore(#280): rename template extension

freddieptf:
This PR fixes #280 by
1. Changing all template extension to `.liquid`
2. Replacing use of deprecated `include` with `render`

PR #291, opened 2025-04-29
Introduce View Models

kennsippell:
Introduces typed view models https://github.com/medic/cht-user-management/issues/270

Example:

This adds ViewModels for anything which is rendered directly by a route.  

This is an example view model in liquid/auth/index.ts:

```
export type AuthFormViewModel = {
  domains: AuthenticationInfo[];
  error?: string;
};

export type AuthViewModel = {
  logo: string;
} & AuthFormViewModel;
```

And the route looks like this:

```
    const viewModel: AuthFormViewModel = {
      domains: Config.getDomains(),
      error
    };
    return resp.view('src/liquid/auth/authentication_form.liquid', viewModel);
```

* Renames html to liquid https://github.com/medic/cht-user-management/issues/280
* Consolidates `include` and `render` to use render for everything
* Moves view and routes from /new and /edit to be under /multiplace/new and /multiplace/edit
* Component which are rendered by routes and not by views are moved out of the components folder
* Add documentation of expected information at the top of all components which don't have a viewmodel


kennsippell:
I didn't see https://github.com/medic/cht-user-management/pull/290 when I approached this work. I'll rebase this to be based off that work.

PR #294, opened 2025-05-12
fix(#159): disable auth checks for sse endpoint

### LINKED ISSUE #159: When cookie expires, requests to /connection/login are spammed rapidly
Screenshot from 2024-04-29 08-26-27
Looks like about 2 or 3 per second

freddieptf:
Fixes #159 

The browser automatically tries to reconnect to the SSE endpoint any time the SSE connection is closed or errs for any reason. In #159, when the cookie expires the connection can not be established successfully because of the authentication check on the server and so the browser will just keep on trying to connect until it succeeds. This PR removes the auth check for the SSE endpoint so the app can connect successfully and since it's an SSE endpoint, the connection will remain idle until the server has something to send to the app.


kennsippell (changes requested):
I don't think making this unauthenticated is the right approach.

1. it has sensitive info and so ensuring that it never leaks anything needs to be avoided carefully. you dont have any tests to avoid a regression
2. the way you've implented this keeps connections open to unauthenticated requests which exposes the service to DOS 

i'd recommend just adding a `Retry-After: ` header requests which fail because they are unauthenticated. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After  It seems this is respected by XHTML

Or you could consider a circuit break pattern?

```
<script>
  (function () {
    let failureCount = 0;
    const maxFailures = 5;

    document.addEventListener("htmx:sseError", function (evt) {
      failureCount++;
      console.warn(`SSE connection failed ${failureCount} times`);
      if (failureCount >= maxFailures) {
        console.log("Max failures reached. Stopping SSE connection.");
        evt.target.removeAttribute("sse-connect");
      }
    });
  })();
</script>
```

freddieptf:
> it has sensitive info and so ensuring that it never leaks anything needs to be avoided carefully. you dont have any tests to avoid a regression

The connection can be established but the server does not send updates to the client if the request is not authenticated. 

> the way you've implented this keeps connections open to unauthenticated requests which exposes the service to DOS

Yeah, i thought it wasn't much of a risk. I went with this approach instead of listening for the `htmx:sseError` event because there's no way to know why the connection failed i.e if it was an authentication error or another random connection error. The circuit breaker approach could break the "update stream" in some cases where retries are valid but i think with a high enough retry count it should be fine

src/public/app.js
```diff
@@ -0,0 +1,11 @@
+(function() {
+	let failureCount = 0;
+	const maxFailures = 5;
+	document.addEventListener("htmx:sseError", function(evt) {
```
kennsippell, line 4: recommend using double space `  ` instead of tabs

src/public/app.js
```diff
@@ -0,0 +1,11 @@
+(function() {
+	let failureCount = 0;
+	const maxFailures = 5;
```
kennsippell, line 3: this is like a true const for readability. recommend a naming pattern like `MAX_FAILURES`?

src/liquid/app/view.liquid
```diff
@@ -56,7 +57,7 @@
     </div>
 
     <script type="text/javascript">
-      bulmaCalendar.attach('input[type='date']', {
+      bulmaCalendar.attach('input[type=\'date\']', {
```
kennsippell, line 60: :P

src/public/app.js
```diff
@@ -0,0 +1,11 @@
...
+	document.addEventListener("htmx:sseError", function(evt) {
+		failureCount++;
+		if (failureCount >= maxFailures) {
+			console.log("Max failures reached. Stopping SSE connection.");
+			evt.target.removeAttribute("sse-connect");
```
kennsippell, line 8: an alternative or additional action here might be to:

1. confirm the user is not authenticated
2. navigate to /login

very optional

kennsippell (approved):

PR #298, opened 2025-05-29
fix(ui): improve required input indicators and fix user role select box (#297)

### LINKED ISSUE #297: Feedback from MoH Capacity Building
We had a recent capacity building session with MoH Togo and some minor feedback were noted:

- [x] The user role select box is empty with no label
- [x] When property is required we should have an `*` suffix on the input lablel
- [x] When property is required we should add a `required` on the input tag
- [x] Update the config.json (add more level in the hierarchy to avoid error)

paulpascal:
This PR fixes #297

The remaining feedback will be fixed in another [PR](https://github.com/medic/cht-user-management/pull/299) since its more related to the config.


freddieptf (approved):
LGTM, please take note of the updated development process https://github.com/medic/cht-user-management#development-1

> Submit a PR, and make sure that the PR title is clear, readable, and follows the strict commit message format described in the commit message format section below. If the PR title does not comply, automatic release will fail

The PR title should at-least summarize what changed

PR #304, opened 2025-05-30
chore(#303): override image tag during deployment

### LINKED ISSUE #303: Deploy action does not deploy the latest version
Deploy action doesn't update to the latest image as the wrong value is overriden in https://github.com/medic/cht-user-management/blob/ba0eb522d7e912adc4ef929c5b5ed6caf1079ffb/.github/workflows/deploy.yml#L66

freddieptf:
Fixes #303

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -63,12 +63,12 @@ jobs:
 
       - name: Upgrade users-chis-civ
         run: |
-          helm upgrade --namespace ${{vars.NAMESPACE}} --values scripts/deploy/values/users-chis-civ.yaml users-chis-civ medic/cht-user-management --set tag=$PKG_VERSION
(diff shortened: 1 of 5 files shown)


Hareet:
From [helm upgrade docs](https://helm.sh/docs/helm/helm_upgrade/), these are the relevant parts for us:

> You can specify the '--values'/'-f' flag multiple times. The priority will be given to the last (right-most) file specified
> You can specify the '--set' flag multiple times. The priority will be given to the last (right-most) set specified.
> You can update the values for an existing release with this command as well via the '--reuse-values' flag. The 'RELEASE' and 'CHART' arguments should be set to the original parameters, and existing values will be merged with any values set via '--values'/'-f' or '--set' flags. Priority is given to new values.

You can handle that a few ways. In the current PR, we are removing image_tag from values and only defining it by set in the cli command. That's okay, as long as all documentation and CI/CD calls reflect that new process. You can also leave image_tag in the values file and override as necessary based on the docs above.

The current PR works, I'll approve it. It's up to you if you want to test the overrides or just consistently document everywhere that this is how you deploy/run.

Hareet (approved):
👍 

Left comment above

PR #319, opened 2025-10-29
fix(#317): use E.164 phone number format

### LINKED ISSUE #317: Fix issue that adds space to a  chp contact phone number
Adding a number like `0720123456` in the `add new chp form` is staged and saved in the instance as `0720 123456` for the contact

Ideally the number should be saved without the space to avoid unintended problems elsewhere. For instance sending sms via rapidpro fails because of that space.

freddieptf:
Fixes #317


Benmuiruri:
My bad. Added to tomorrow's to-do list.

PR #323, opened 2026-02-06
fix(#322): set reduce param to false for view query

### LINKED ISSUE #322: Search fails on instances on CHT v5
Search has been failing on instances running cht v5 or greater. The view being relied on for search `contacts_by_type` has a reduce function on CHT v5, meaning the `reduce` param needs to be explicitly set to `false` in the search query to get results

freddieptf:
Fixes #322


Benmuiruri (approved):
LGTM

PR #326, opened 2026-03-26
fix(#325): show Nepali characters

### LINKED ISSUE #325: Place names in Nepali (Devanagari script) not showing up in the hierarchy dropdowns
Place names with Nepali (Devanagari script) characters are not shown in the hierarchy dropdown suggestions.

Examples:
-

I have several health centers with Nepali characters:

Image

However, only those in English/Latin scripts are shown in the CHT user management:

Image

Image

I tried with partial Englsh and Nepali names, and only English part is shown. Nepali characters are stripped.

binokaryg:
closes #325
- Made regex validator more flexible to support Unicode
- Added a test

### DIFF
--- a/src/validation/validator-string.ts
+++ b/src/validation/validator-string.ts
@@ -6,7 +6,7 @@ export default class ValidatorString implements IValidator {
   }
 
   format(input : string) : string {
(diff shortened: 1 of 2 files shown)


freddieptf (approved):
LGTM

Issue #332, opened 2026-05-06
Login authentication should support providing plain-name domains


kennsippell:
The CHW-R is integrating with UMT.  CHW-R does know that a CHP is in a specific county (eg. Nairobi) but doesn't know the eCHIS URL for that domain (eg. nairobi.echis.go.ke).  It also shouldn't need this information.

API endpoints like #330 need an auth token and post login payloads to `/authenticate` including the domain URL. Presently, CHW-R would need to know the URL of all eCHIS instances.  This issue tracks an improvement where login is powered by county names instead of URLs.

PR #333, opened 2026-05-06
feat(#332): authentication payload should use friendly domain names

kennsippell:
#332 

1. Changes login to be via friendly domain instead of hostname
2. Changes the spelling of counties to match what is in eCHIS Registry
3. Updates/adds Kenya training instance data to config


paulpascal (approved):
Looks Good To Me  👍🏻

Issue #335, opened 2026-05-13
New API to support login to CHT instances with SSO enabled


kennsippell:
CHT 4.20 enabled SSO. UMT should support authentication for users via SSO. 

This is useful for the MoH Kenya CHW Registry project. In particular, that scenario only needs a Server-to-Server solution and doesn't require SSO users to login through the UMT UI.

PR #331, opened 2026-05-06
feat(#330): add new /api endpoints for CHW-Registry

kennsippell:
Adds
* `/api/v1/search` - Fuzzy search for contacts
* `/api/v1/create` - Create new users 
* `/api/v1/manage-hierarchy` - Move or merge contacts

#330

PR #336, opened 2026-05-13
feat(#335): add new endpoint /api/v1/sso-login

kennsippell:
#335

Adds /api/v1/sso-login

1. Find an OIDC provider and add a user with email like `rsa@registry.local`
2. Configure CHT for SSO. Details in https://docs.communityhealthtoolkit.org/hosting/sso
3. Add a CHT user with user management permissions. Set their OIDC email to the same as step 1
4. Get an access token from your OIDC. Something like:
  ```
  ACCESS_TOKEN=$(curl -s -X POST 'http://127.0.0.1:54321/auth/v1/token?grant_type=password' \
    -H 'Content-Type: application/json' \
    -H 'apikey: x' \
    -d '{"email":"rsa@registry.local","password":"pwd"}' | jq -r .access_token)
  ```
5. Drive the SSO dance through UMT
```
curl -i -X POST http://localhost:3500/api/v1/sso-login \
  -H 'Content-Type: application/json' \
  -d "$(jq -n --arg t "$ACCESS_TOKEN" '{domain: "$Development", access_token: $t}')"
```

This is the third of three cascading PRs: #333 , #331 should merge before this as setup 
Thanks to @jkuester for helping to give this PR direction


src/routes/sso-login.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Config } from '../config';
+
+export default async function ssoLoginRoute(fastify: FastifyInstance) {
+  fastify.post('/api/v1/sso-login', async (req) => {
+    const formBody: any = req.body;
+    ensureJsonObjectBody(formBody);
+
+    const { domain, access_token } = formBody;
+    if (!domain || !access_token) {
+      throw AuthError.MISSING_CREDENTIALS();
```
jkuester: This could result in some confusion since the message for this error is:

> Missing username or password

src/lib/sso-login.ts
```diff
@@ -0,0 +1,136 @@
...
+
+function serializeCookies(jar: CookieJar, host: string): string {
+  const bucket = jar.get(host);
+  if (!bucket) {
+    return '';
+  }
+  return Array.from(bucket.entries()).map(([n, v]) => `${n}=${v}`).join('; ');
+}
+
+async function follow(start: string, accessToken: string, maxHops = 12): Promise<{ jar: CookieJar; finalUrl: string }> {
```
jkuester: Minor, but it is odd to have maxHops as a parameter with a default value that is never overridden.  Might be more clear to just use a top-level `MAX_REDIRECT_HOPS` const.

src/lib/sso-login.ts
```diff
@@ -0,0 +1,136 @@
...
+  const jar: CookieJar = new Map();
+  let current = start;
+  for (let hop = 0; hop < maxHops; hop++) {
+    const url = new URL(current);
+    const cookies = serializeCookies(jar, url.host);
+    const headers: Record<string, string> = {
+      Authorization: `Bearer ${accessToken}`,
```
jkuester: I guess this will work as it, but it does make me nervous to just always be blindly passing this accessToken for each redirect. If I understand things correctly, we need to send this header when the CHT instance redirects to the OIDC provider (so we authenticate with the OIDC provider).  However, we should not need to send it on the initail call to `/medic/login/oidc/authorize`. It is essentially leaking the OIDC accessToken to the CHT server.  

Passing it to _all_ the other later redirects also feels like a security risk since a malicious actor might find a way to slip in a redirect to a URL they control and then steal the accessToken.

I guess the most straightforward way to harden this would be to have a specific `redirect_url` pattern that the URL must match before we actually pass the accessToken.  (Either configure that as part of the `AuthenticationInfo`, or even just accept it as a body property on the `/api/v1/sso-login` request.)

src/lib/sso-login.ts
```diff
@@ -0,0 +1,136 @@
...
+        continue;
+      }
+    }
+    if (res.status >= 200 && res.status < 300) {
+      return { jar, finalUrl: current };
+    }
+    const errBody = typeof res.data === 'string' ? res.data.slice(0, 200) : '';
+    throw new Error(`SSO dance failed at ${current}: HTTP ${res.status} ${errBody}`);
+  }
+  throw new Error(`SSO dance exceeded ${maxHops} hops`);
```
jkuester: Might be helpful to include `current` here too (so we can see what the last hop was).

src/lib/sso-login.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    if (res.status >= 300 && res.status < 400) {
+      let loc = res.headers.location as string | undefined;
+      // CHT 5.1.2 emits 302 with the redirect URL in the body, not the Location header.
+      const body = typeof res.data === 'string' ? res.data : '';
+      if (!loc && /^https?:\/\//.test(body.trim())) {
+        loc = body.trim();
+      }
+      if (loc) {
+        current = loc.startsWith('http') ? loc : new URL(loc, current).toString();
```
jkuester:
```suggestion
        current = new URL(loc, current).toString();
```

I might be missing something here, but I thought the URL constructor worked fine with `http`.... 🤔

src/lib/sso-login.ts
jkuester: 😅 Really, this whole function is begging to be re-written as a recursive function....

src/lib/sso-login.ts
```diff
@@ -0,0 +1,136 @@
...
+ *   `AuthSession` on the callback hop.
+ */
+export async function ssoLogin(authInfo: AuthenticationInfo, accessToken: string): Promise<SsoLoginResult> {
+  const chtBaseUrl = `${authInfo.useHttp ? 'http' : 'https'}://${authInfo.domain}`;
+  const start = `${chtBaseUrl}/medic/login/oidc/authorize`;
+  const { jar, finalUrl } = await follow(start, accessToken);
+
+  // CHT may redirect across hostnames mid-chain (e.g. localhost <-> LAN IP).
+  // Pick whichever bucket actually carries the session cookie.
+  const bucket = findBucketWith(jar, COUCH_AUTH_COOKIE_NAME);
```
jkuester: This makes me a bit nervious.  I would have expected the Couch AuthSession cookie to always end up in the bucket for `authInfo.domain`.  Were you seeing different behavior?

src/routes/api.ts
jkuester, line 1: FTR, I have not reviewed this file since I think it was covered in https://github.com/medic/cht-user-management/pull/331.

jkuester (approved):
LGTM!  🚀 This is great and pretty much exactly what I was hoping for after our discussion!  It is really nice to see how you were able to cleanly slot it next to the default session handling so pretty much everything post-sso-login just flows as normal.

I left a few minor suggestions and one comment with some more significant considerations around protecting the access_token.  Since the primary targeted deployment for this functionality owns everything from the code consuming this endpoint (in the registry), to the UMT deployment, to the CHT instances, to the OIDC provider it seems the practical implications of this are minimal, but I do feel that the best practice would be to treat the access_token as highly sensitive.

src/lib/sso-login.ts
kennsippell: i think this was just a scenario when testing on localhost and the domain changed to a resolved local IP. perhaps I should revert that and just not support scenarios where the hsotname changes for any reason (probably just have to test on 127.0.0.1 instead of localhost)

src/lib/sso-login.ts
kennsippell: definite yes to the const. 
medium feelings on the recursion 🖌️

jkuester: :sweat_smile: Yeah, please do not rewrite this just to be recursive!  I just got a chuckle reading through it because passing this param is exactly how I would have done it for a recursive function.

src/lib/sso-login.ts
jkuester: Ahhh, :grimacing: yeah, that makes sense.  I guess the main risk here is accidentally ended up with a non-CHT `AuthSession` cookie coming from one of the other redirects... :shrug:   Probably not worth worrying about in this first iteration.

src/routes/sso-login.ts
kennsippell: Yep. Added unique error message.

src/routes/api.ts
kennsippell: Good skip. Reverted

src/lib/sso-login.ts
kennsippell: :scissors: Cut it

src/lib/sso-login.ts
kennsippell: I created an allowed list of IDP providers within the configuration system.

kennsippell:
@jkuester I know you signed off on this already, but I made quite a few changes in response to your comments and I'd love a quick check again before merge. Even just a "go ahead" nod.

kennsippell:
Locally tested, internationally regretted

PR #338, opened 2026-05-23
New endpoint /api/v1/cancel-job 

kennsippell:
Adds a new endpoing `/api/v1/cancel-job` which allows for the cancelation of a move or merge job. Used for "roll back scenarios" in Kenya's CHW Registry.


kennsippell:
Taking a different path

PR #342, opened 2026-05-29
fix(#340): show validation error tooltip on invalid badge for select_one fields

### LINKED ISSUE #340: Importing new users from CSV, the Require password change field is not validated
1. Create csv for import as below
2. Import it

Observed: CHW users show as "invalid" with no tooltip
Expected: validation errors should show a clear message

Affected CHU | Sub County | CHU Name | CHU Code | Link Facility Name | Link Facility MFL Code | CHA Name | CHA Phone | Require password change
-- | -- | -- | -- | -- | -- | -- | -- | --
  | Narok West | Lemek Community Health Unit | 111113 | Lemek Community Health Unit | 11113 | Lemek Community Health Unit CHA | 712341236 | No

In this case, the `Require password change` should be `no` instead of `No`

vikrantwiz02:
## Problem

When importing a CSV with `Require password change` = `No` (capital N), the row is marked **invalid** but no tooltip appears on the badge — leaving users with no explanation of what went wrong.

Two overlapping bugs caused this:
1. The `invalid` status badge only attached a `data-tooltip` for `uploadError`, not for `validationErrors`
2. A CSS typo `has-tooltipl-multiline` (5 occurrences) was preventing multiline tooltip styling everywhere

Additionally, `ValidatorSelectOne` returned a generic `false` on failure, producing the unhelpful message `"Invalid value selected"` instead of telling the user what values are valid.

## Changes

- **`validator-select-one.ts`**: return `"Valid choices are: yes, no"` instead of `false`, so the error message names the accepted values
- **`validator-select-multiple.ts`**: update filter to `!== true` to correctly handle the new string return from `select-one`
- **`place.ts`**: add `validationErrorSummary` as an own enumerable getter via `Object.defineProperty` in the constructor, so it always reflects the current state of `validationErrors` and is visible to LiquidJS templates
- **`place_item.liquid`**: attach `data-tooltip` with validation errors to the `invalid` badge; fix 3 CSS typos
- **`list_cell.liquid`** + **`place_list.liquid`**: fix CSS typo `has-tooltipl-multiline` → `has-tooltip-multiline`
- **`test/validation.spec.ts`**: assert new descriptive error messages for `select_one` invalid scenarios

## Testing

- Import CSV with `Require password change` = `No` → hover the **invalid** badge → tooltip reads `Valid choices are: yes, no`
- Fix to `no` (lowercase) → row becomes valid and ready to upload
- All 209 unit tests pass

### DIFF
--- a/src/services/place.ts
+++ b/src/services/place.ts
@@ -253,6 +253,10 @@ export default class Place {
     return Object.keys(this.validationErrors as any).length > 0;
   }
 
+  public get validationErrorSummary(): string {
+    return Object.values(this.validationErrors || {}).join('\n');
+  }
+
   public get isDependant() : boolean {
     return !!this.resolvedHierarchy.find(hierarchy => hierarchy?.type === 'local') || this.hasSharedUser;
   }
--- a/src/services/place.ts
+++ b/src/services/place.ts
@@ -51,7 +51,7 @@ export default class Place {
   public state : PlaceUploadState;
 
   public validationErrors?: { [key: string]: string };
-  public validationErrorSummary: string = '';
+  declare validationErrorSummary: string;
   public uploadError? : string;
 
   constructor(type: ContactType) {
@@ -64,6 +64,11 @@ export default class Place {
     this.resolvedHierarchy = [];
     this.warnings = [];
     this.userRoleProperties = {};
+    Object.defineProperty(this, 'validationErrorSummary', {
+      enumerable: true,
+      configurable: true,
+      get: () => Object.values(this.validationErrors ?? {}).join('\n'),
+    });
   }
 
   /*
@@ -217,7 +222,6 @@ export default class Place {
     for (const property of propertiesWithErrors) {
       this.validationErrors[property.propertyNameWithPrefix] = property.validationError as string;
     }
-    this.validationErrorSummary = Object.values(this.validationErrors).join('\n');
   }
 
   public generateUsername(): string {
--- a/src/services/place.ts
+++ b/src/services/place.ts
@@ -51,6 +51,7 @@ export default class Place {
   public state : PlaceUploadState;
 
   public validationErrors?: { [key: string]: string };
+  public validationErrorSummary: string = '';
   public uploadError? : string;
 
   constructor(type: ContactType) {
@@ -216,6 +217,7 @@ export default class Place {
     for (const property of propertiesWithErrors) {
       this.validationErrors[property.propertyNameWithPrefix] = property.validationError as string;
     }
+    this.validationErrorSummary = Object.values(this.validationErrors).join('\n');
   }
 
   public generateUsername(): string {
@@ -251,10 +253,6 @@ export default class Place {
 
   public get hasValidationErrors() : boolean {
     return Object.keys(this.validationErrors as any).length > 0;
-  }
-
-  public get validationErrorSummary(): string {
-    return Object.values(this.validationErrors || {}).join('\n');
   }
 
   public get isDependant() : boolean {
--- a/src/validation/validator-select-multiple.ts
+++ b/src/validation/validator-select-multiple.ts
@@ -16,11 +16,11 @@ export default class ValidatorSelectMultiple implements IValidator {
     const selectOneValidator = new ValidatorSelectOne();
     const stringValidator = new ValidatorString();
 
-    const selectedValues = this.parseInput(input, stringValidator); 
+    const selectedValues = this.parseInput(input, stringValidator);
     const invalidValues = selectedValues.filter(
       value => selectOneValidator.isValid(value, property) !== true
     );
-
+    
     if (invalidValues.length > 0) {
       return `Invalid values for property "${property.friendly_name}": ${invalidValues.join(', ')}`;
     }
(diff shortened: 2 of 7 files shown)


vikrantwiz02:
Hey @kennsippell, I've raised a PR with the fix for this. Sorry for direct PR, without an official assignment! It's ready for your review whenever you have time.

kennsippell (changes requested):
Thanks for taking the time to work this out! Looks great. I'm seeing a small issue when I do the following:

1. Create a CSV for importing Kenyan CHUs (see below) but with invalid line "Require password change = No"
2. Import the csv
3. Look at tooltip for invalid row

image

CSV:
Affected CHU | Sub County | CHU Name | CHU Code | Link Facility Name | Link Facility MFL Code | CHA Name | CHA Phone | Require password change
-- | -- | -- | -- | -- | -- | -- | -- | --
  | Narok West | Lemek Community Health Unit | 111113 | Lemek Community Health Unit | 11113 | Lemek Community Health Unit CHA | 712341236 | No

test/validation.spec.ts
```diff
@@ -80,9 +80,9 @@ const scenarios: Scenario[] = [
   { type: 'select_one', prop: undefined, isValid: false, error: 'Required', formatted: '' },
   { type: 'select_one', prop: ' male', isValid: true, propertyParameter: GENDER_OPTIONS },
   { type: 'select_one', prop: 'female ', isValid: true, propertyParameter: GENDER_OPTIONS },
-  { type: 'select_one', prop: 'FeMale ', isValid: false, propertyParameter: GENDER_OPTIONS },
-  { type: 'select_one', prop: 'f', isValid: false, propertyParameter: GENDER_OPTIONS },
-  { type: 'select_one', prop: '', isValid: false, propertyParameter: GENDER_OPTIONS },
+  { type: 'select_one', prop: 'FeMale ', isValid: false, propertyParameter: GENDER_OPTIONS, error: 'Valid choices' },
+  { type: 'select_one', prop: 'f', isValid: false, propertyParameter: GENDER_OPTIONS, error: 'Valid choices' },
```
kennsippell, line 85: possible to add a test for the exact case called out in the issue? So `Require password change = No` and the validationError that results?

src/validation/validator-select-multiple.ts
```diff
@@ -18,7 +18,7 @@ export default class ValidatorSelectMultiple implements IValidator {
 
     const selectedValues = this.parseInput(input, stringValidator); 
     const invalidValues = selectedValues.filter(
-      value => !selectOneValidator.isValid(value, property)
+      value => selectOneValidator.isValid(value, property) !== true
```
kennsippell, line 21: can you explain this change? when would we want a falsy output from isValid to now be selected here?

vikrantwiz02: Actually, before this PR, `ValidatorSelectOne.isValid` returned a boolean. This PR updates it to return an error string on failure (e.g. `"Valid choices are: yes, no"`).

Bcauz a non-empty string is truthy, the old `!isValid` check would incorrectly evaluate to `false` and allow invalid values to pass through. Changing it to `!== true` ensures we correctly catch both the old `false` returns and the new error strings.

test/validation.spec.ts
vikrantwiz02: Done! Added `YES_NO_OPTIONS = { yes: 'Yes', no: 'No' }` to mirror the exact config and a test case for `prop: 'No'` that asserts the full error message `'Valid choices are: yes, no'`.

vikrantwiz02:
@kennsippell, empty tooltip was caused by `validationErrorSummary` being a prototype getter. LiquidJS only resolves own enumerable properties, not prototype getters. Converted it to a plain property computed inside `validate()` after `validationErrors` is populated.

src/services/place.ts
```diff
@@ -216,6 +217,7 @@ export default class Place {
     for (const property of propertiesWithErrors) {
       this.validationErrors[property.propertyNameWithPrefix] = property.validationError as string;
     }
+    this.validationErrorSummary = Object.values(this.validationErrors).join('\n');
```
kennsippell: ValidationErrors can be written to, so it isn't safe to just set this once during construction. What if the validation errors change?

Should this be a getter?

vikrantwiz02: You're right @kennsippell, that's a valid concern. If `validationErrors` is ever mutated after `validate()` runs, the summary would go stale.

The tricky part is we can't just convert it back to a class getter, LiquidJS only resolves own enumerable properties, not prototype getters, which was the original reason we moved away from one.

The fix that satisfies both: use `Object.defineProperty` on the instance inside the constructor. That creates an **own enumerable getter** (not a prototype getter), so LiquidJS can see it and it always reflects the current state of `validationErrors`:

```ts
// constructor
Object.defineProperty(this, 'validationErrorSummary', {
  enumerable: true,
  configurable: true,
  get: () => Object.values(this.validationErrors ?? {}).join('\n'),
});
```

And swap the class field declaration to just `declare validationErrorSummary: string;` so TypeScript is happy but no initializer is emitted. Then the manual assignment in `validate()` can be removed entirely.

Will push the update.

PR #343, opened 2026-06-02
feat(#355): additional API endpoints for CHW-R integration

kennsippell:
* Adds `/api/v1/create-user` - create CHA users in eCHIS when created in the registry
* Adds `/api/v1/disable-users-at` - to disable CHP users on exit
* Adds `/api/v1/set-user-facilities` - to reassign CHUs between CHAs
* Adds a `?clear_cache=1` param to `/api/v1/search`
* Renames `/api/v1/create` to `/api/v1/create-user-and-place` to disambiguate
* Resolves a caching bug in which places created via `/api/v1/create-user-and-place` can't be found via search without busting cache

### DIFF
--- a/src/routes/api.ts
+++ b/src/routes/api.ts
@@ -2,7 +2,7 @@ import _ from 'lodash';
 
 import { ChtApi } from '../lib/cht-api';
 import { Config } from '../config';
-import { FastifyInstance } from 'fastify';
+import { FastifyInstance, FastifyRequest } from 'fastify';
 import Fuse from 'fuse.js';
 import Place, { PlaceUploadState } from '../services/place';
 import PlaceFactory from '../services/place-factory';
@@ -13,6 +13,7 @@ import RemotePlaceResolver from '../lib/remote-place-resolver';
 import WarningSystem from '../warnings';
 import { DisableUsers } from '../lib/disable-users';
 import { SetUserFacilities } from '../services/set-user-facilities';
+import { OidcUserPayload } from '../services/oidc-user-payload';
 
 const DEFAULT_THRESHOLD = 0.6;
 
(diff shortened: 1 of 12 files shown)


Benmuiruri:
@paulpascal let me review this .

src/routes/api.ts
```diff
@@ -64,54 +67,100 @@ export default async function api(fastify: FastifyInstance) {
...
-      .search(formBody[baseHierarchyLevel.property_name] ?? '')
-      .map(({ item, score }) => ({
-        name: item.name.original,
-        place_id: item.id,
-        score: score ?? 1,
-      }))
-      .sort((a, b) => a.score - b.score);
-
-    return hits;
+    const disabled = await DisableUsers.disableUsersAt([facility.place_id], chtApi);
```
Benmuiruri, line 140: The only suggestion is to match the others and add an accompanying test

```
    try {
      const disabled = await DisableUsers.disableUsersAt([facility.place_id], chtApi);
      return {
        place_id: facility.place_id,
        place_name: facility.name,
        disabled,
      };
    } catch (e: any) {
      return { error: e.response?.data?.error?.message ?? e.toString() };
    }
```

Benmuiruri (approved):
LGTM! Just one suggestion.

src/routes/api.ts
kennsippell: Great feedback! Done

kennsippell:
A MR for CHW-R which depends on `/api/v1/disable-users-at` has merged (https://gitlab.echis.go.ke/echis-support-apps/chw-registry-frontend/-/merge_requests/100). I haven't done e2e testing on `/api/v1/create-user` and `/api/v1/set-user-facilities` because those areas of CHW-R are not yet complete. I'm debating removing those endpoints from this PR and adding them in a different change, but since they are reviewed and tested I think it makes sense to commit as-is. Worst case, they will need to be changed but they currentliy have no consumers.

PR #350, opened 2026-06-12
fix(#349): add chw-registry-dev to idpOrigins

### LINKED ISSUE #349: Bearer Missing on CHW-R Dev Instance during SSO operations
image

This error is shown in CHW-R because `https://chw-registry-dev.test.co.ke` isn't in the list of idpOrigins.

kennsippell:
Closes #349

### DIFF
--- a/src/config/chis-ke/config.json
+++ b/src/config/chis-ke/config.json
@@ -210,8 +210,9 @@
     }
   ],
   "idpOrigins": [
+    "https://chpregistry.echis.go.ke",
     "https://chpregistry-staging.echis.go.ke",
-    "https://chpregistry.echis.go.ke"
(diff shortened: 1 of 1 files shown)


src/config/chis-ke/config.json
```diff
@@ -210,8 +210,9 @@
     }
   ],
   "idpOrigins": [
+    "https://chpregistry.echis.go.ke",
```
karimkkanji, line 213: Let's also add this one since it is what we'll use for production eventually
```suggestion
    "https://chpregistry.echis.go.ke",
    "https://chwregistry.echis.go.ke",
```

eljhkrr (approved):
Looks good, thanks for the feedback Karim

Issue #355, opened 2026-06-17
Additional requirements for API with CHW-R


kennsippell:
1. Endpoint to create CHA users in eCHIS when created in the registry
2. Endpoint to disable CHP users when they exit
3. Endpoint to reassign CHUs between CHAs
4. Add ?clear_cache=1 param to /api/v1/search

PR #360, opened 2026-06-19
fix(#355): design adjustments to API endpoints during CHA creation in CHW-R

kennsippell:
* `/api/v1/create-user`
  1. Create new primary contact to represent the new user
  2. When creating a CHA, the facilities should be exclusively owned by the new user. Add a flag `exclusiveFacilities`  unassigns facilities from other users.
* `/api/v1/set-user-facilities` - change interface to accept the  `oidc_username` to streamline fix for https://github.com/medic/cht-core/issues/11129 (when ready)
* Removes `/api/v1/create` (backward compat alias for `/api/v1/create-user-and-place`)
* Forgot to commit some code review feedback from #343

#355


kennsippell:
**Scenario**. User A manages two facilities, A1 and A2. They were provisioned with a single, shared primary contact P — one person doc that is the primary contact of both A1 and A2, and physically lives under A1. Now I want to reassign A1 to a new User B.

Here's everything that seems to be required:

  1. Create B's contact — POST /api/v1/people to create a person under A1.
  2. Remove A1 from User A — POST /api/v1/users/{A} with place: [A2].
  3. Deal with P, the displaced contact. This is the hard part. After step 2, P still lives under A1 (which A no longer manages), is still the primary contact of A2 (which A does manage), and is User A's own user.contact. So
  I have to:
    - look up where P is still a primary contact (medic/_design/medic/_view/contacts_by_primary_contact),
    - if it's still primary of a place A manages (A2), move P there (edit its parent),
    - if nothing references it and no user points at it, it's an orphan and can be deleted — but if a user still references it, I can't safely delete it at all.
  4. Repoint A1 — POST /api/v1/places/{A1} to set A1's primary contact to B's contact.
  5. Create User B — POST /api/v1/users with B's contact and facility_id: [A1].

And if I ever reassign User A back into A1 (or anywhere new), and P was deleted as an orphan along the way, I have to recreate it first.


Today this PR doesn't do a good job of handling P, the displaced primary contact. I'm unclear if I should file a bug, handle all this complexity, or just ignore it (as it has been ignored by the `/reassign` endpoint).

kennsippell:
The above scenario should be fully resolved now

Benmuiruri:
Hi @kennsippell , I'll review this today

src/routes/api.ts
```diff
@@ -229,14 +246,49 @@ function validateCreateUser(
...
+    if (e?.response?.status === 404) {
+      throw new Error(`Facility place "${facilityId}" was not found in this eCHIS instance`);
+    }
+    throw e;
+  }
+}
+
+// Always creates a fresh person from the supplied contact details and makes it the primary contact of
+// every facility passed
+async function createPrimaryContactForFacilities(
```
Benmuiruri, line 276: Is there a possibility of  creating the contact, set contac as primary on every facility but `createUser` may fail. Should we log the `contactId` on failure ? 

Or perhaps change the loop to be create the person, create the user, then run the updatePlace loop.

src/routes/api.ts
```diff
@@ -66,28 +67,38 @@ export default async function api(fastify: FastifyInstance) {
...
+        : undefined;
+
+      return {
+        success: true,
+        username: payload.username,
+        ...(unassigned ? { unassigned } : {}),
+      };
     } catch (e: any) {
-      return { error: e.response?.data?.error?.message ?? e.toString() };
+      return { error: e.response?.data?.error?.message ?? e.message ?? e.toString() };
```
Benmuiruri, line 101: Minor: Match L157 and L177 to this cleaner `e.message`

Benmuiruri (approved):
LGTM

One suggestion and an open question about order of operations in the createuser

src/routes/api.ts
kennsippell: Nice catch

src/routes/api.ts
kennsippell: Yes this is a great idea.

1. Reordered the operations as you suggested (create person, create user, update place loop)
2. Return the person's id if it is created and other steps fail
3. The update place loop attempts to update all places even if failures happen

PR #362, opened 2026-06-22
fix(#361): improve performance of search operations

### LINKED ISSUE #361: Search functionality is slow, frequently taking over 30 seconds to complete
Investigate why the `/search` and `/api/v1/search` operations are so slow in production.

1. Login to https://users.echis.go.ke using Training-1
2. Do a search with warm cache

Expected: Fast
Observed: 30seconds+

kennsippell:
Closes #361 

PR| I (cold) | II (warm) | III (warm)
-- | -- | -- | --
Before | 46s | 30s | 28s
After | 25s | <1s | <1s

Regression from https://github.com/medic/cht-user-management/commit/631975dfc06d815d1fc4e9133a50fd613d5bf2a4

### DIFF
--- a/src/lib/remote-place-cache.ts
+++ b/src/lib/remote-place-cache.ts
@@ -36,6 +36,7 @@ export default class RemotePlaceCache {
   private static readonly CACHE_CHECK_PERIOD = 120;
   private static cache: NodeCache = new NodeCache({
     stdTTL: this.getCacheTTL(),
+    useClones: false,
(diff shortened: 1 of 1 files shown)


kennsippell:
@paulpascal @freddieptf @Benmuiruri Could one of you review this in the next few days? It is important because  problematic performance is actually causing `/_healthz` probes to fail in K8 and forcing restarts of the production https://users.echis.go.ke pod

Benmuiruri (approved):

PR #367, opened 2026-06-25
chore(#368): enforce unit tests typing

### LINKED ISSUE #368: Enable typscript typing for tests
### Description
Currently types are not enforced in unit test causing random bugs and subtle misalignment between code and tests

jonathanbataire:
fixes #368

### DIFF
--- a/test/integration/manage-hierarchy.spec.ts
+++ b/test/integration/manage-hierarchy.spec.ts
@@ -62,10 +62,10 @@ describe('integration/manage-hierarchy',  function () {
   const sessionCache = SessionCache.getForSession(session);
   const chtApi = () => mockChtApi(
     [
-      { id: 'from-sub', name: 'From Sub', lineage: [], type: 'remote' },
-      { id: 'to-sub', name: 'To Sub', lineage: [], type: 'remote' }
+      { _id: 'from-sub', name: 'From Sub', lineage: [], type: 'remote' },
+      { _id: 'to-sub', name: 'To Sub', lineage: [], type: 'remote' }
     ],
-    [{ id: 'chu-id', name: 'c-h-u', lineage: ['from-sub'], type: 'remote' }],
+    [{ _id: 'chu-id', name: 'c-h-u', lineage: ['from-sub'], type: 'remote' }],
   );
 
   it('should process a move contact job', async () => {
@@ -76,7 +76,7 @@ describe('integration/manage-hierarchy',  function () {
     encodeTokenStub.returns('encoded-token');
     decodeTokenStub.returns(session);
 
-    await MoveLib.scheduleJob(
+    await (MoveLib as any).scheduleJob(
       formData, contactType, sessionCache, chtApi(), moveContactQueue
     );
 
@@ -107,7 +107,7 @@ describe('integration/manage-hierarchy',  function () {
     encodeTokenStub.returns('encoded-token');
     decodeTokenStub.throws(new Error('Missing WORKER_PRIVATE_KEY'));
 
-    await MoveLib.scheduleJob(
+    await (MoveLib as any).scheduleJob(
       formData, contactType, sessionCache, chtApi(), moveContactQueue
     );
 
--- a/test/lib/authentication.spec.ts
+++ b/test/lib/authentication.spec.ts
@@ -31,14 +31,14 @@ describe('lib/authentication/authentication.ts', () => {
 
   it('invalid session cannot be decoded for cookie', () => {
     const session = mockChtSession();
-    delete session.username;
+    delete (session as any).username;
     const encoded = Auth.encodeTokenForCookie(session);
     expect(() => Auth.createCookieSession(encoded)).to.throw('invalid CHT session information');
   });
 
   it('invalid session cannot be decoded for workers', () => {
     const session = mockChtSession();
-    delete session.username;
+    delete (session as any).username;
     const encoded = Auth.encodeTokenForWorker(session);
     expect(() => Auth.createWorkerSession(encoded)).to.throw('invalid CHT session information');
   });
--- a/test/lib/authentication.spec.ts
+++ b/test/lib/authentication.spec.ts
@@ -30,15 +30,23 @@ describe('lib/authentication/authentication.ts', () => {
   });
 
   it('invalid session cannot be decoded for cookie', () => {
-    const session = mockChtSession();
-    delete (session as any).username;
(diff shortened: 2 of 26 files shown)


test/integration/manage-hierarchy.spec.ts
```diff
@@ -76,7 +76,7 @@ describe('integration/manage-hierarchy',  function () {
     encodeTokenStub.returns('encoded-token');
     decodeTokenStub.returns(session);
 
-    await MoveLib.scheduleJob(
+    await (MoveLib as any).scheduleJob(
```
freddieptf, line 79: not sure about this one chief, do we really need this when calling static funcs?

test/lib/authentication.spec.ts
```diff
@@ -31,14 +31,14 @@ describe('lib/authentication/authentication.ts', () => {
 
   it('invalid session cannot be decoded for cookie', () => {
     const session = mockChtSession();
-    delete session.username;
+    delete (session as any).username;
```
freddieptf: should this be `ChtSession` instead?

test/lib/disable-users.spec.ts
```diff
@@ -17,7 +18,7 @@ describe('lib/disable-users.ts', () => {
       username: 'user',
       place: { _id: PLACE_ID }
     }]);
-    const actual = await DisableUsers.disableUsersAt([PLACE_ID], cht);
+    const actual = await DisableUsers.disableUsersAt([PLACE_ID], cht as unknown as ChtApi);
```
freddieptf: `cht as unknown as ChtApi` is this some ts quirk you're working around?

Image

test/integration/manage-hierarchy.spec.ts
jonathanbataire: :)

jonathanbataire: im mainly just adding types to whats existing without fiddling with existing logic
the `any` changes nothing compared to what was there
how can we handle this better?

test/lib/disable-users.spec.ts
jonathanbataire: the problem is mainly how these tests were written originally 
no types meaning you could write anything that doesn't fit the the argument type and things work
`cht` is not type CHTApi the double type casting `as unkown as ChtApi` is a way to force `cht` be type `ChtApi` even though it doesnt fit the original type
these mistakes were caused by lack of type enforcing

test/lib/authentication.spec.ts
jonathanbataire: the properties `username` and `password` of `CHTSession` are readonly 
meaning they cant be deleted 
this test is deleting them 
in order for that to happen we need to cast it to `any`

test/lib/disable-users.spec.ts
jonathanbataire: I have made improvements to CHTApi mock logic to reduce this double casting

test/integration/manage-hierarchy.spec.ts
freddieptf: in this case i find it odd that a call needs some type, i assumed `await MoveLib.scheduleJob(...` would just work

test/lib/authentication.spec.ts
freddieptf: makes sense, what about initializing an invalid cht session to avoid the deletion?

test/integration/manage-hierarchy.spec.ts
jonathanbataire: movelib.schedulejob takes only 2 args this test is passing 5 
hence the casting

test/lib/authentication.spec.ts
jonathanbataire: i have improved this

test/integration/manage-hierarchy.spec.ts
freddieptf: yeah that seems wrong and i don't think these integration tests run at all. I think it can be handled in another PR. Could you create another issue for this?

test/lib/sso-login.spec.ts
```diff
@@ -94,9 +94,9 @@ describe('lib/sso-login.ts', () => {
       // The dance "succeeds" (mocked) but we just want to verify no header leaked.
       await ssoLogin(mockAuthInfo, ACCESS_TOKEN, IDP_ORIGINS).catch(() => undefined);
 
-      const attackerCall = stub.getCalls().find(c => new URL(c.args[0].url).host === 'idp.example.com.attacker.com');
+      const attackerCall = stub.getCalls().find(c => new URL(c.args[0].url as string).host === 'idp.example.com.attacker.com');
       expect(attackerCall, 'attacker host was visited').to.exist;
-      expect(attackerCall!.args[0].headers.Authorization).to.be.undefined;
+      expect((attackerCall!.args[0].headers as any)?.Authorization).to.be.undefined;
```
freddieptf: is there some `http.Headers` type you can use here?

test/integration/manage-hierarchy.spec.ts
jonathanbataire: https://github.com/medic/cht-user-management/issues/384

test/worker/cht-conf-worker.spec.ts
```diff
@@ -32,7 +33,7 @@ describe('worker/cht-conf-worker', () => {
   describe('handleJob', () => {
     it('move contacts', async () => {
       const job = mockJob('move');
-      await ChtConfWorker.handleJob(job);
+      await (ChtConfWorker as any).handleJob(job as unknown as Job<any, any, string>);
```
freddieptf, line 36: so something about doing this cast pattern, i think it's hiding some actual issues with these tests. most functions being tested on this `ChtConfWorker` class are private in the first place and i think this should fail

freddieptf: i guess it's fine if the only concern is if `executeStub` is called

jonathanbataire: id say its revealing the flaws rather than hiding them
also slapping private on a method wont make it private at runtime (only checked at compiler) thats why these tests work even before this update
introduction of types has no effect on this the `as any` silences the `private` compile error
true js private would be `#methodName()`

freddieptf (approved):
i'd say this is a good first step, need to have a deeper look at some of this tests

PR #371, opened 2026-06-30
fix(#364): refine fuzzy name matching

### LINKED ISSUE #364: Capture and refine matches for /api/v1/search
* Capture a list of matches and non-matches within the context of the CHW-Registry
* Update endpoint to accept the first, middle, and last names separately to match on permutations

cc @Benmuiruri as FYI

kennsippell:
* Switches name matching algorithm from Fuse.js to levenshtein
* Tunes threshold parameter based on real-world examples from #364 (training) 
* Captures positive and negative cases for CHW-R name matching in tests so requirements can become clearer and algorithm can learn and grow
* Removes threshold parameter control from CHW-R. UMT owns the search verdict.

Of the 8 scenarios captured in #364, only 1 of them passed on `main`. All 8 of them are now passing in this branch.

Closes #364

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,6 @@
         "fastify": "^4.27.0",
         "fastify-metrics": "^11.0.0",
         "fastify-sse-v2": "^3.1.2",
-        "fuse.js": "^7.3.0",
         "html-minifier": "^4.0.0",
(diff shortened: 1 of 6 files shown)


Benmuiruri (approved):
LGTM 

Thanks for this much needed enhancement.

PR #374, opened 2026-07-13
fix(#372): add chwregistry prod url for chis-ke

Benmuiruri:
## What

Empty commit to trigger a `semantic-release` version bump (expected `v2.12.3`).

## Why

The chwregistry prod URL config change already landed on `main` in #372, but that commit used a non-conventional `UPD:` message. 

This carries a conventional `fix(#372):` message so the release fires and the change gets a proper changelog entry. The diff is intentionally empty.,


jonathanbataire (commented):
@Benmuiruri i dont see any changes in this PR

jonathanbataire (approved):

PR #377, opened 2026-07-13
fix(#375): update echis area name stripping

### LINKED ISSUE #375: Refine matching after apostrophe
I came across Vihiga county user who's matching was failing because their name on registry is `Benjamin Kiarie` and their area is `Benjamin Kiarie's Community Health Volunteer Area` 

Update the stripping to not just Area, but anything coming after the apostrophe.

Benmuiruri:
Community Health Volunteer areas stored as `<name>'s Community Health Volunteer Area` failed to match the registry search name, because the chis-ke strip rule only removed " Area" — leaving the possessive 's, which pushed the fuzzy score over the match threshold.

Changes:
- `src/config/chis-ke/config.json `— CHV-area replacement property now also strips the possessive 's and trailing descriptor  before matching. Intra-name apostrophes (e.g. Ng'ang'a) are preserved.

Closes #375

### DIFF
--- a/test/lib/name-match.spec.ts
+++ b/test/lib/name-match.spec.ts
@@ -15,6 +15,8 @@ describe('lib/name-match.ts', () => {
     { search: 'Christine Wanyaga Muchiri',   echis: 'Christine Muchiri Wanyaga Area', match: true },
     { search: 'CHARLES Barkasau BARGASAU',   echis: 'CHARLES BARGASAU Area',          match: true },
     { search: 'MARGARET CHEMOIWO JEPSERGON', echis: 'MARGARET JEPSERGON Area',        match: true },
+    { search: 'Benjamin Kiarie Muiruri',     echis: "Benjamin Kiarie's Community Health Volunteer Area", match: true },
+    { search: "James Ng'ang'a Kamau",        echis: "James Ng'ang'a's Community Health Volunteer Area",  match: true },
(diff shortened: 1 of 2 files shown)


karimkkanji:
Thanks @Benmuiruri. This is very critical to ensure we can match the CHPs to their areas as they could potentially have different naming.

test/lib/name-match.spec.ts
```diff
@@ -15,6 +15,8 @@ describe('lib/name-match.ts', () => {
     { search: 'Christine Wanyaga Muchiri',   echis: 'Christine Muchiri Wanyaga Area', match: true },
     { search: 'CHARLES Barkasau BARGASAU',   echis: 'CHARLES BARGASAU Area',          match: true },
     { search: 'MARGARET CHEMOIWO JEPSERGON', echis: 'MARGARET JEPSERGON Area',        match: true },
+    { search: 'Benjamin Kiarie Muiruri',     echis: "Benjamin Kiarie's Community Health Volunteer Area", match: true },
+    { search: "James Ng'ang'a Kamau",        echis: "James Ng'ang'a's Community Health Volunteer Area",  match: true },
```
kennsippell, line 19: Can we consider a test case that confirms the space. Confirming that a name like `James Ng'sang` is not affected.

kennsippell (approved):
Nice

PR #378, opened 2026-07-13
fix(#376): don't return external sources if contact type has no external source mapping

### LINKED ISSUE #376: External source returns blank page if contact type has no config
### Description
Having 2 contact types:
-  one with external source mapping, call it "A" 
- another without external source mapping, call it "B" 

This means the one without external source mapping in config ("B") has external source button active and returns a blank page
### expected behavior
Contact type "B" shouldn't show external source button and not allow searching external source because it has nothing in config

Example below parish has no external source mapping yet button appears and search page is blank

Image

Image

jonathanbataire:
### Changes
- If contact type has no external source mapping return []
- adds external config tests
fixes #376

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -125,6 +125,10 @@ export class Config {
     return contactMatch;
   }
 
+  public static contactTypes(): ContactType[] {
+    return config.contact_types;
+  }
+
   public static getParentProperty(contactType: ContactType): HierarchyConstraint {
     const parentMatch = contactType.hierarchy.find(c => c.level === 1);
     if (!parentMatch) {
@@ -262,10 +266,6 @@ export class Config {
       ...contactType.hierarchy
     ]
       .some(prop => prop.external_mapping?.[sourceId]?.is_filter);
-  }
-
-  public static contactTypes(): ContactType[] {
-    return config.contact_types;
   }
 
   // get all external source parameters from configuration
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -117,16 +117,16 @@ const { config } = partnerConfig;
 export class Config {
   private constructor() { }
 
+  public static contactTypes(): ContactType[] {
+    return config.contact_types;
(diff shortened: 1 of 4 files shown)


src/config/index.ts
```diff
@@ -117,26 +117,10 @@ const { config } = partnerConfig;
 export class Config {
   private constructor() { }
 
-  public static getExternalSources(id?: string): ExternalSource[] {
```
jonathanbataire, line 120: moved below for better code organization

src/config/index.ts
```diff
@@ -145,38 +129,6 @@ export class Config {
     return contactMatch;
   }
 
-  // get all external source parameters from configuration
-  public static getExternalSourceConfigById(sourceId: string, contactTypeName: string): ExternalSourceConfig {
```
jonathanbataire, line 149: moved below for better code organization

Benmuiruri (approved):
lGTM , just one comment

jonathanbataire:
> lGTM , just one comment

cant find the comment @Benmuiruri did you add it?

src/routes/add-place.ts
```diff
@@ -93,6 +88,10 @@ export default async function addPlace(fastify: FastifyInstance) {
     if (!source) {
       return renderError(`Unkown external source ${sourceId}`);
     }
+    
+    if (Object.keys(searchParams).length === 0) {
```
Benmuiruri, line 92: The form at `search_external_source.liquid:10` always submits a hidden `place_type`, can searchParams be empty ?

Benmuiruri (commented):
I had not added it

src/routes/add-place.ts
jonathanbataire: this is an API safe guard /validation
cases where data is sent but not thru the form

Benmuiruri: cool

